PR #2, opened 2022-06-23
Add fourier methods and tests

Aman-Pandey-afk:
Added the functions from fourier.py, along with required functions from gti.py and util.py (Remaining will be implemented according to use). Tests are thoroughly covered. DataFrames and HDF5 support for large data also implemented.


test/test_fourier.jl
```diff
@@ -0,0 +1,407 @@
...
+            times,
+            gti,
+            segment_size,
+            dt,
+            norm=norm,
+            use_common_mean=false,
+            silent=true,
+            fluxes=nothing,
+        ).power
+        @test std(out_comm)≈std(out) rtol=0.1
```
giordano, line 174: Relative tolerance of 10% sounds fairly bad.  Is this the precision within which the two functions are supposed to agree?

test/test_fourier.jl
```diff
@@ -0,0 +1,407 @@
...
+
+    s_discard = Symbol.(discard)
+
+    for key in propertynames(table1)
+        if key in s_discard
+            continue
+        end
+        oe, oc = getproperty(table1,key), getproperty(table1,key)
+        if oe isa Integer || oe isa String
+            @assert oe==oc
```
giordano, line 11: [`@assert`](https://docs.julialang.org/en/v1/base/base/#Base.@assert) is supposed to be used only for debugging purposes.  Please read the docstrings which strongly warns against any other uses.

test/test_fourier.jl
```diff
@@ -0,0 +1,407 @@
...
+            dt,
+            norm=norm,
+            use_common_mean=use_common_mean,
+            silent=true,
+            fluxes1=counts,
+            fluxes2=counts2,
+            errors1=errs,
+            errors2=errs2,
+        )
+        discard = [m for m in propertynames(out_ev) if m == Symbol("variance")]
```
giordano, line 310: Simpler:
```suggestion
        discard = [m for m in propertynames(out_ev) if m == :variance]
```

test/test_fourier.jl
```diff
@@ -0,0 +1,407 @@
...
+    end
+end
+
+@testset "test_norm" begin
+    mean = var = 100000
+    N = 1000000
+    dt = 0.2
+    meanrate = mean / dt
+    lc = rand(Poisson(mean),N)
+    pds = abs.(fft(lc)).^2
```
giordano, line 326:
```suggestion
    pds = abs2.(fft(lc))
```

test/test_fourier.jl
```diff
@@ -0,0 +1,407 @@
...
+    pds = (abs.(fft(lc)) .^ 2)[good]
+    lc_bksub = lc .- mean
+    pds_bksub = (abs.(fft(lc_bksub)) .^ 2)[good]
+    lc_renorm = lc / mean
+    pds_renorm = (abs.(fft(lc_renorm)) .^ 2)[good]
+    lc_renorm_bksub = lc_renorm .- 1
+    pds_renorm_bksub = (abs.(fft(lc_renorm_bksub)) .^ 2)[good]
+
+    @testset "test_leahy_bksub_var_vs_standard" begin
+        """Test that the Leahy norm. does not change with background-subtracted lcs"""
```
giordano, line 356: This string is doing nothing.  What is is supposed to do?  A comment?

test/test_gti.jl
```diff
@@ -0,0 +1,44 @@
+@testset "test_time_intervals_from_gtis" begin
+    """Test the division of start and end times to calculate spectra."""
```
giordano, line 2: Other extra strings.

test/test_gti.jl
```diff
@@ -0,0 +1,44 @@
...
+        start_times, stop_times = time_intervals_from_gtis([[0 400]; [1022 1200];
+                                      [1210 1220];], 128)
+        @test start_times == [0, 128, 256, 1022]
+        @test stop_times == start_times .+ 128
```
giordano, line 6: These lines have an off indentation

test/test_gti.jl
```diff
@@ -0,0 +1,44 @@
...
+end
+
+@testset "test_time_intervals_from_gtis_frac" begin
+    """Test the division of start and end times to calculate spectra."""
+        start_times, stop_times = time_intervals_from_gtis([[0 400]; [1022 1200];
+                                      [1210 1220];], 128, fraction_step=0.5)
+        @test start_times == [0, 64, 128, 192, 256, 1022]
+        @test stop_times == start_times .+ 128
```
giordano, line 14: Same comments as above.

src/fourier.jl
```diff
@@ -0,0 +1,755 @@
...
+        end
+        @yield cts
+    end
+end
+
+function avg_pds_from_iterable(flux_iterable, dt; norm="frac", use_common_mean=true,
+                               silent=false)
+    local_show_progress = show_progress
+    if silent
+        local_show_progress = (a)->a
```
giordano, line 224: [`identity`](https://docs.julialang.org/en/v1/base/base/#Base.identity)
```suggestion
        local_show_progress = identity
```

src/fourier.jl
```diff
@@ -0,0 +1,755 @@
...
+
+function raw_coherence(cross_power, power1, power2, power1_noise, power2_noise,
+                       n_ave; intrinsic_coherence=1.0)
+    bsq = bias_term(power1, power2, power1_noise, power2_noise, n_ave;
+                    intrinsic_coherence=intrinsic_coherence)
+    num = real(cross_power .* conj(cross_power)) - bsq
+    if num isa Array{T} where T<:Any
+        num[num .< 0] = real(cross_power .* conj(cross_power))[num .< 0]
+    elseif num < 0
+        num = real(cross_power * conj(cross_power))
```
giordano, line 94: Are you sure this line is doing what you think?  Please elaborate in words what's your expectation.

src/fourier.jl
```diff
@@ -0,0 +1,755 @@
...
+    n_ave = 0
+    freq = nothing
+
+    sum_of_photons = 0
+    common_variance = nothing
+    fgt0 = nothing
+    n_bin = nothing
+
+    for flux in local_show_progress(flux_iterable)
+        if isnothing(flux) || all(flux .== 0)
```
giordano, line 238:
```suggestion
        if isnothing(flux) || all(iszero, flux)
```
has two advantages: you don't allocate the array `flux .== 0`, and also `all` is able to return early at the first non-zero element of `flux`.

src/fourier.jl
```diff
@@ -0,0 +1,755 @@
...
+function avg_cs_from_iterables_quick(flux_iterable1,flux_iterable2,
+                                    dt; norm="frac")
+    unnorm_cross = unnorm_pds1 = unnorm_pds2 = nothing
+    n_ave = 0
+    fgt0 = n_bin = freq = nothing
+
+    sum_of_photons1 = sum_of_photons2 = 0
+
+    for (flux1, flux2) in zip(flux_iterable1, flux_iterable2)
+        if isnothing(flux1) || isnothing(flux2) || all(flux1 .== 0) || all(flux2 .== 0)
```
giordano, line 348:
```suggestion
        if isnothing(flux1) || isnothing(flux2) || all(iszero, flux1) || all(iszero, flux2)
```

src/fourier.jl
```diff
@@ -0,0 +1,755 @@
...
+    cross = unnorm_cross = unnorm_pds1 = unnorm_pds2 = pds1 = pds2 = nothing
+    n_ave = 0
+    fgt0 = n_bin = freq = nothing
+
+    sum_of_photons1 = sum_of_photons2 = 0
+    common_variance1 = common_variance2 = common_variance = nothing
+
+    for (flux1, flux2) in local_show_progress(zip(flux_iterable1,
+                                                flux_iterable2))
+        if isnothing(flux1) || isnothing(flux2) || all(flux1 == 0) || all(flux2 == 0)
```
giordano, line 464:
```suggestion
        if isnothing(flux1) || isnothing(flux2) || all(iszero, flux1) || all(iszero, flux2)
```

src/fourier.jl
```diff
@@ -0,0 +1,755 @@
...
+
+    flux_iterable1 = get_flux_iterable_from_segments(
+        times1, gti, segment_size; n_bin, fluxes=fluxes1, errors=errors1
+    )
+    flux_iterable2 = get_flux_iterable_from_segments(
+        times2, gti, segment_size; n_bin, fluxes=fluxes2, errors=errors2
+    )
+
+    is_events = all([isnothing(val) for val in (fluxes1, fluxes2, errors1,
+                                                errors2)])
```
giordano, line 723:
```suggestion
    is_events = all(isnothing, (fluxes1, fluxes2, errors1,errors2))
```

giordano (changes requested):
The main problem is that the code feels like written in Python, not in Julia.  It's very unidiomatic.  Function signatures are completely non-annotated, lots of keyword arguments with `nothing` as default value, that's really not a thing in Julia.

Also, I haven't actually run the code, but I'm certain performance isn't optimal, there are plenty of type-instabilities, did you use `@code_warntype`?

src/utils.jl
```diff
@@ -0,0 +1,6 @@
+function sum_if_not_none_or_initialize(A,B)
+    if isnothing(A) 
+        return deepcopy(copy((B)))
+    end
+    return A + B
+end
```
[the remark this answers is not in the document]
Aman-Pandey-afk: I think I have used copy() in vain, but the function is executed in loop, so I will need a deepcopy of B when A is nothing (otherwise B points to A).

test/test_fourier.jl
Aman-Pandey-afk: I thought of providing only one tests for each compare_tables execution, so asserted the conditions and finally tested for true, if I test for every key value, there are around 500 tests related to them.

test/test_fourier.jl
Aman-Pandey-afk: Yes it was comment :)

src/utils.jl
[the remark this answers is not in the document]
giordano: > so I will need a deepcopy of B when A is nothing (otherwise B points to A).

What does this mean?

src/fourier.jl
Aman-Pandey-afk: I can't use dot broadcast. The num .<0 returns a boolean array (suppose B). B[i] is false if num[i]<0 and num[B] gives a new array only consisting of those elements of num where B is true. (I think filter could be used to create the array, but dot broadcasting not in my opinion).

src/utils.jl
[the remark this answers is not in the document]
Aman-Pandey-afk: Basically it is a recursive copy of objects, but here it is just to ensure that we don't pass a reference of A to B. Shallow copy could be used, but I though there must be some motive in using deepcopy in python too (like maybe variance could be an array in some case so shallow copy doesn't guarantee that A and B are independent).

src/fourier.jl
```diff
@@ -0,0 +1,755 @@
+function positive_fft_bins(n_bin; include_zero = false)
+    minbin = 2
+    if include_zero
+        minbin = 1
+    end
+
+    if n_bin%2 == 0
+        return (minbin : floor(Int, n_bin / 2))
+    end
+    return (minbin : floor(Int, (n_bin+1)/ 2))
```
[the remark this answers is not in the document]
Aman-Pandey-afk: No, I think convert could be used instead of floor or maybe even concise n_bin÷2 if even and (n_bin+1)÷2 if odd.

src/utils.jl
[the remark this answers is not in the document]
giordano: Specifically I'm missing what

> (otherwise B points to A)

means. Maybe a concrete reproducible example would help.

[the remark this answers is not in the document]
Aman-Pandey-afk: The specific use-case is here: https://github.com/StingraySoftware/stingray/blob/1bd574a284dfe60177e3313d50cc5700f4d1ec42/stingray/fourier.py#L1005-L1007
As you can see if I just use `sum_if_not_none_or_initialize(::Nothing,B) = B` , it will point to the reference of B. Ex:
image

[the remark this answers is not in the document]
giordano: As a general remark, never show screenshots, whoever wants to help you can't copy from that, moreover if someone wants to come back to this discussion whatever is written in the screenshot doesn't show up in search. They aren't really much useful.

That said, I'm not sure that's a problem in practice here.

Also, I think this discussion is a bit out of the point now: I believe you should first understand how to avoid the overuse of nothing and write more idiomatic Julia code here, at which point a new solution may arise.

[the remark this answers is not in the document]
matteobachetti: Hi @giordano, @Aman-Pandey-afk  : 
This is the use case: in another function, I am calculating (e.g.) an average of many periodograms produced over a number of short time series. In each loop, I produce a periodogram that has to be summed to the total periodogram. The summed periodogram is initialized as `None`. In the first loop, this function will find that the total periodogram is None, and initialize it to a deep copy of the first periodogram. From that point on, all subsequent arrays will be summed to the total periodogram in place.
The reason for the deep copy is to avoid side effects on the first periodogram, in case it is useful later.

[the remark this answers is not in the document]
giordano: Ok, but the `deepcopy` is a red herring here, my main point is that we should move away from overusing `nothing`.  There _are_ definitely good use cases for `nothing` in Julia, but I feel like all the uses here come from the fact it was used (for good reasons!) in Python, not that it's necessary in Julia.

src/fourier.jl
giordano: My point is that you can define a function (please use a more sensible name than "foo")
```julia
foo(num::Real) = num < 0 ? real(cross_power * conj(cross_power))
```
and then here you do
```julia
num .= foo.(num)
```
which would work whatever `num` is, either a scalar number or an aray.  Much simpler, and it doesn't allocate lots of temporary new arrays.

giordano:
BTW, another thing I just realised is that in the code there are a lot of `all(arr .== 0)` (which [I suggested](https://github.com/matteobachetti/Stingray.jl/pull/2#discussion_r905554233) to replace with `all(iszero, arr)` because it's faster and it avoids allocation of temporary arrays), but with floating point numbers it isn't generally a good idea to check exact equality with 0, you'd use for example [`isapprox`](https://docs.julialang.org/en/v1/base/math/#Base.isapprox) (with a non-zero absolute tolerance).  Is exact equality with zero really what you want to check?

src/fourier.jl
Aman-Pandey-afk: Problem is cross power is also an array, so this way won't work. If we have to use function we need to pass index too, in that case also, there is a separation between array and number. num is simply iterable of floats (I think it works with tuples too in python)

giordano: Why it wouldn't work?

Aman-Pandey-afk:
I have used ≈ wherever required, the places you have suggested to use `iszero` requires the array sum to be not equal to zero (leading to unnecessary calculations) thus, exact equality is needed to be checked. I will use `iszero` instead of `flux1 .== 0`.

src/fourier.jl
Aman-Pandey-afk:
```
julia> cross_power = [1+im,1+2im,2]
3-element Vector{Complex{Int64}}:
 1 + 1im
 1 + 2im
 2 + 0im

julia> num=[-1,-2,3]
3-element Vector{Int64}:
 -1
 -2
  3

julia> foo(num::Real) = num < 0 ? real(cross_power .* conj(cross_power)) : num   
foo (generic function with 1 method)

julia> num .= foo.(num)
ERROR: MethodError: Cannot `convert` an object of type Vector{Int64} to an object of type Int64
```

giordano:
```julia
julia> foo(num::Real, cross_power::Complex) = num < 0 ? real(cross_power * conj(cross_power)) : num
foo (generic function with 1 method)

julia> cross_power = [1 + im, 1 + 2im, 2]
3-element Vector{Complex{Int64}}:
 1 + 1im
 1 + 2im
 2 + 0im

julia> num = [-1.0, -2.0, 3.0]
3-element Vector{Float64}:
 -1.0
 -2.0
  3.0

julia> num .= foo.(num, cross_power)
3-element Vector{Float64}:
 2.0
 5.0
 3.0
```

Aman-Pandey-afk: Oh yes we can do that, I looked now upon different ways to broadcast arguments in functions, nice!

src/fourier.jl
```diff
@@ -0,0 +1,755 @@
...
+    two_n_ave = 2 * n_ave
+    if common_ref
+        Gsq = (cross_power * conj(cross_power)).real
+        bsq = bias_term(seg_power, ref_power, seg_power_noise, ref_power_noise,
+                        n_ave)
+        frac = (Gsq - bsq) / (ref_power - ref_power_noise)
+        power_over_2n = ref_power / two_n_ave
+
+        # Eq. 18
+        dRe = dIm = dG = sqrt(power_over_2n * (seg_power - frac))
```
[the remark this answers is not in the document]
Aman-Pandey-afk: Will fix this by today, have to develop test for it.

src/gti.jl
```diff
@@ -0,0 +1,98 @@
...
+end
+
+function calculate_segment_bin_start(startbin::Integer, stopbin::Integer,
+                                     nbin::Integer; fraction_step::Real=1)
+    st = floor.(range(startbin, stopbin, step=Int64(nbin * fraction_step)))
+    if st[end] == stopbin
+        st = st[1:end-1]
+    end
+    if st[end] + nbin > stopbin
+        return st[1:end-1]
```
giordano, line 26: I think this is equivalent to
```suggestion
        pop!(st)
    end
    if st[end] + nbin > stopbin
        pop!(st)
```
right?

Also,  I [have already mentioned](https://github.com/matteobachetti/Stingray.jl/pull/2#discussion_r905500295) that slicing creates copies, so in `st = st[1:end-1]` you are creating a copy of `st[1:end-1]` before reassigning it to the `st` identifier.  [`pop!`](https://docs.julialang.org/en/v1/base/collections/#Base.pop!) simply removes the last element of the collection, without assuming anything about indexing of the array (Julia is 1-based by default, but arbitrary indices are allowed).

src/gti.jl
```diff
@@ -0,0 +1,98 @@
...
+        dt = Statistics.median(diff(time))
+    end
+
+    epsilon_times_dt = epsilon * dt
+    nbin = Int64(round(segment_size / dt))
+
+    spectrum_start_bins = Int64[]
+
+    gti_low = (@view gtis[:, 1]) .+ (dt / 2 - epsilon_times_dt)
+    gti_up = (@view gtis[:, 2]) .- (dt / 2 - epsilon_times_dt)
```
giordano, line 44:
```suggestion
    gti_low = @view(gtis[:, 1]) .+ (dt ./ 2 .- epsilon_times_dt)
    gti_up = @view(gtis[:, 2]) .- (dt ./ 2 .- epsilon_times_dt)
```

src/gti.jl
```diff
@@ -0,0 +1,98 @@
...
+
+        newtimes = range(g1, g2 - segment_size + epsilon, step=Float64(segment_size) * fraction_step)
+        append!(spectrum_start_times,newtimes)
+    end
+    return spectrum_start_times, spectrum_start_times .+ segment_size
+end
+
+function calculate_segment_bin_start(startbin::Integer, stopbin::Integer,
+                                     nbin::Integer; fraction_step::Real=1)
+    st = floor.(range(startbin, stopbin, step=Int64(nbin * fraction_step)))
```
giordano, line 21: Unless you have a specific reason for using 64-bit integers, it's usually more natural to use `Int`, which is the integer type with the same number of bits as the current architecture (32-bit on 32-bit systems, 64-bit on 64-bit integers).  `Int` is also the type of integer literals in Julia
```suggestion
    st = floor.(range(startbin, stopbin, step=Int(nbin * fraction_step)))
```

src/gti.jl
```diff
@@ -0,0 +1,98 @@
...
+
+function bin_intervals_from_gtis(gtis::AbstractMatrix{<:Real}, segment_size::Real,
+                                 time::AbstractVector{<:Real}; dt=nothing, 
+                                 fraction_step::Real=1, epsilon::Real=0.001)
+    if isnothing(dt)
+        dt = Statistics.median(diff(time))
+    end
+
+    epsilon_times_dt = epsilon * dt
+    nbin = Int64(round(segment_size / dt))
```
giordano, line 39:
```suggestion
    nbin = round(Int, segment_size / dt)
```

src/gti.jl
```diff
@@ -0,0 +1,98 @@
...
+
+    spectrum_start_bins = Int64[]
```
giordano, line 41:
```suggestion
    spectrum_start_bins = Int[]
```

test/test_fourier.jl
```diff
@@ -0,0 +1,419 @@
...
+        end
+    end
+    @test test_result
+end
+
+@testset "positive_fft_bins" begin
+    freq = fftfreq(11)
+    good = freq .> 0
+    goodbins = positive_fft_bins(11)
+    @test freq[good]==freq[goodbins]
```
giordano, line 50: In Julia masks are less frequently used than in Python, [`filter`](https://docs.julialang.org/en/v1/base/collections/#Base.filter)-ing is much more common:
```suggestion
    goodbins = positive_fft_bins(11)
    @test filter(x -> x >0, freq) == freq[goodbins]
```

test/test_fourier.jl
```diff
@@ -0,0 +1,419 @@
...
+    data2 = rand.(Poisson.(data))
+    ft1 = fft(data1)
+    ft2 = fft(data2)
+    dt = 0.01
+    N = length(data)
+    mean = Statistics.mean(data)
+    meanrate = mean / dt
+    freq = fftfreq(length(data), 1/dt)
+    good = (freq .> 0 .&& freq .< 0.1)
+    ft1, ft2 = ft1[good], ft2[good]
```
giordano, line 69: This is more idiomatic:
```suggestion
    good = 0 .< freq .< 0.1
    keepat!(ft1, good)
    keepat!(ft2, good)
```

test/test_fourier.jl
```diff
@@ -0,0 +1,419 @@
...
+    pds2 = normalize_periodograms(
+        ft2 .* conj(ft2), dt, N, mean_flux = mean, norm="abs", power_type="real")
+
+    p1noise = poisson_level("abs",meanrate=meanrate)
+    p2noise = poisson_level("abs",meanrate=meanrate)
+
+    @testset "test_intrinsic_coherence" begin
+        coh = estimate_intrinsic_coherence(
+            cross, pds1, pds2, p1noise, p2noise, N)
+        @test all(.≈(coh, 1, atol=0.001))
```
giordano, line 83: Same comment as https://github.com/matteobachetti/Stingray.jl/pull/2#discussion_r905554233
```suggestion
        @test all(x -> isapprox(x, 1; atol=0.001), coh)
```
Side note: I _**strongly**_ recommend always using `;` to introduce keyword arguments.

test/test_fourier.jl
```diff
@@ -0,0 +1,419 @@
...
+    end
+
+    @testset "test_raw_high_coherence" begin
+        coh = raw_coherence(cross, pds1, pds2, p1noise, p2noise, N)
+        @test all(.≈(coh, 1, atol=0.001))
```
giordano, line 88: Again, same comment as https://github.com/matteobachetti/Stingray.jl/pull/2#discussion_r905554233
```suggestion
        @test all(x -> isapprox(x, 1; atol=0.001), coh)
```

test/test_fourier.jl
```diff
@@ -0,0 +1,419 @@
...
+    end
+
+    @testset "test_raw_low_coherence" begin
+        nbins = 2
+        C, P1, P2 = cross[1:nbins], pds1[1:nbins], pds2[1:nbins]
+        bsq = bias_term(P1, P2, p1noise, p2noise, N)
+        # must be lower than bsq!
+        low_coh_cross = rand.(Normal.(bsq.^0.5 / 10, bsq.^0.5 / 100)) .+ 0.0im
+        coh = raw_coherence(low_coh_cross, P1, P2, p1noise, p2noise, N)
+        @test all(coh .== 0)
```
giordano, line 98: Again (3), same comment as https://github.com/matteobachetti/Stingray.jl/pull/2#discussion_r905554233
```suggestion
        @test all(iszero, coh)
```

test/test_fourier.jl
```diff
@@ -0,0 +1,419 @@
...
+        coh = raw_coherence(cross, pds1, pds2, p1noise, p2noise, N)
+        @test all(.≈(coh, 1, atol=0.001))
+    end
+
+    @testset "test_raw_low_coherence" begin
+        nbins = 2
+        C, P1, P2 = cross[1:nbins], pds1[1:nbins], pds2[1:nbins]
+        bsq = bias_term(P1, P2, p1noise, p2noise, N)
+        # must be lower than bsq!
+        low_coh_cross = rand.(Normal.(bsq.^0.5 / 10, bsq.^0.5 / 100)) .+ 0.0im
```
giordano, line 96:
```suggestion
        low_coh_cross = rand.(Normal.(bsq.^0.5 ./ 10, bsq.^0.5 ./ 100)) .+ 0.0im
```

test/test_fourier.jl
```diff
@@ -0,0 +1,419 @@
...
+        @test coh_sngl==real(C .* conj(C))[1] / (P1[1] * P2[1])
+    end
+    
+end
+
+@testset "test_fourier" begin
+    dt = 1
+    len = 100
+    ctrate = 10000
+    N = Int64(len / dt)
```
giordano, line 120: Same function already suggested in https://github.com/matteobachetti/Stingray.jl/pull/2#discussion_r905451223:
```suggestion
    N = len ÷ dt
```

test/test_fourier.jl
```diff
@@ -0,0 +1,419 @@
...
+    segment_size = 20.0
+    times2 = sort(rand(Uniform(0, len), len * ctrate))
+    counts2 = fit(Histogram, times2, bins).weights
+    errs2 = fill!(similar(counts2),1) * sqrt(ctrate)
+
+    @test get_average_ctrate(times, gti, segment_size) == ctrate
+    @test get_average_ctrate(bin_times, gti, segment_size; counts=counts) == ctrate
+
+    @testset "test_fts_from_segments_cts_and_events_are_equal" begin
+        N = Int64(round(segment_size / dt))
```
giordano, line 137:
```suggestion
        N = segment_size ÷ dt
```

src/gti.jl
giordano: But actually, integer division (already suggested in https://github.com/matteobachetti/Stingray.jl/pull/2#discussion_r905451223) is much idiomatic
```suggestion
    nbin = segment_size ÷ dt
```

test/test_fourier.jl
```diff
@@ -0,0 +1,419 @@
...
+        fts_evts = [
+            f for f in get_flux_iterable_from_segments(times, gti, segment_size, n_bin=N)
+        ]
```
giordano, line 140: Wouldn't this be equivalent to
```suggestion
        fts_evts = collect(get_flux_iterable_from_segments(times, gti, segment_size; n_bin=N))
```
?  Same below

test/test_fourier.jl
```diff
@@ -0,0 +1,419 @@
...
+        fts_cts = [
+            f
+            for f in get_flux_iterable_from_segments(
+                bin_times, gti, segment_size, fluxes=counts
+            )
+        ]
+        for (fe, fc) in zip(fts_evts, fts_cts)
+            @test fe==fc
+        end
```
giordano, line 149: Wouldn't this be equivalent to
```suggestion
        @test fts_evts == fts_cts
```

test/test_fourier.jl
```diff
@@ -0,0 +1,419 @@
...
+
+@testset "test_norm" begin
+    mean = var = 100000
+    N = 1000000
+    dt = 0.2
+    meanrate = mean / dt
+    lc = rand(Poisson(mean),N)
+    pds = abs2.(fft(lc))
+    freq = fftfreq(N, dt)
+    good = 2:floor(Int,N/2)
```
giordano, line 341: This is always integer division:
```suggestion
    good = 2:(N ÷2)
```

test/test_fourier.jl
```diff
@@ -0,0 +1,419 @@
...
+
+    pdsabs = normalize_abs(pds, dt, size(lc,1))
+    pdsfrac = normalize_frac(pds, dt, size(lc,1), mean)
+    pois_abs = poisson_level("abs", meanrate=meanrate)
+    pois_frac = poisson_level("frac", meanrate=meanrate)
+
+    @test Statistics.mean(pdsabs[good])≈pois_abs rtol=0.02
+    @test Statistics.mean(pdsfrac[good])≈pois_frac rtol=0.02
+
+    mean = var = 100000.
```
giordano, line 351:
```suggestion
    mean = var = 100000.0
```

src/fourier.jl
```diff
@@ -0,0 +1,781 @@
...
+    if norm == "abs"
+        return 2.0 * meanrate
+    elseif norm == "frac"
+        return 2.0 / (meanrate - backrate)^2 * meanrate
+    elseif norm == "leahy"
+        return 2.0
+    elseif norm == "none"
+        return float(n_ph)
+    else 
+        throw(ArgumentError("Incorrect Arguments"))
```
giordano, line 19: Please use a more descriptive error message.  Since this is an `if` on the value of `norm`, the only incorrect argument is `norm`, right?  It'd also be useful to print it out explicitly, and test also the error paths.

src/fourier.jl
```diff
@@ -0,0 +1,781 @@
...
+        power = unnorm_power * 2. * dt / ((mean_flux - background_flux) ^ 2 *
+                                          n_bin)
+    else
+        # Note: this corresponds to eq. 3 in Uttley+14
+        power = unnorm_power * 2. * dt / (mean_flux ^ 2 * n_bin)
+    end
+    return power
+end
+
+normalize_abs(unnorm_power:: AbstractVector{<:Number}, dt::Real, n_bin::Integer) =
```
giordano, line 35:
```suggestion
normalize_abs(unnorm_power::AbstractVector{<:Number}, dt::Real, n_bin::Integer) = 
```

src/fourier.jl
```diff
@@ -0,0 +1,781 @@
...
+                                          n_bin)
+    else
+        # Note: this corresponds to eq. 3 in Uttley+14
+        power = unnorm_power * 2. * dt / (mean_flux ^ 2 * n_bin)
+    end
+    return power
+end
+
+normalize_abs(unnorm_power:: AbstractVector{<:Number}, dt::Real, n_bin::Integer) = 
+unnorm_power * 2.0 / n_bin / dt
```
giordano, line 36: Indentation is off:
```suggestion
    unnorm_power * 2 / n_bin / dt
```

src/fourier.jl
```diff
@@ -0,0 +1,781 @@
...
+    elseif norm == "leahy"
+        return 2.0
+    elseif norm == "none"
+        return float(n_ph)
+    else 
+        throw(ArgumentError("Incorrect Arguments"))
+    end
+end
+
+function normalize_frac(unnorm_power, dt::Real, n_bin::Integer,
```
giordano, line 23: What's `unnorm_power`?

src/fourier.jl
```diff
@@ -0,0 +1,781 @@
...
+        return float(n_ph)
+    else 
+        throw(ArgumentError("Incorrect Arguments"))
+    end
+end
+
+function normalize_frac(unnorm_power, dt::Real, n_bin::Integer, 
+                        mean_flux::Real; background_flux::Real=0.0)
+    if background_flux > 0
+        power = unnorm_power * 2. * dt / ((mean_flux - background_flux) ^ 2 *
```
giordano, line 26:
```suggestion
        power = unnorm_power * 2 * dt / ((mean_flux - background_flux) ^ 2 *
```

src/fourier.jl
```diff
@@ -0,0 +1,781 @@
...
+                                          n_bin)
+    else
+        # Note: this corresponds to eq. 3 in Uttley+14
+        power = unnorm_power * 2. * dt / (mean_flux ^ 2 * n_bin)
```
giordano, line 30:
```suggestion
        power = unnorm_power * 2 * dt / (mean_flux ^ 2 * n_bin)
```

src/fourier.jl
```diff
@@ -0,0 +1,781 @@
...
+    end
+    return power
+end
+
+normalize_abs(unnorm_power:: AbstractVector{<:Number}, dt::Real, n_bin::Integer) = 
+unnorm_power * 2.0 / n_bin / dt
+
+normalize_leahy_from_variance(unnorm_power:: AbstractVector{<:Number}, 
+                              variance::Real, n_bin::Integer) = 
+                                unnorm_power * 2.0 / (variance * n_bin)
```
giordano, line 40:
```suggestion
normalize_leahy_from_variance(unnorm_power::AbstractVector{<:Number}, 
                              variance::Real, n_bin::Integer) =
    unnorm_power * 2 / (variance * n_bin)
```

src/fourier.jl
```diff
@@ -0,0 +1,781 @@
...
+
+normalize_leahy_poisson(unnorm_power:: AbstractVector{<:Number}, 
+                        n_ph::Real) = unnorm_power * 2.0 / n_ph
```
giordano, line 43:
```suggestion
normalize_leahy_poisson(unnorm_power::AbstractVector{<:Number}, n_ph::Real) =
    unnorm_power * 2 / n_ph
```

src/fourier.jl
```diff
@@ -0,0 +1,781 @@
...
+    elseif norm == "frac"
+        pds = normalize_frac(
+            unnorm_power, dt, n_bin, mean_flux,
+            background_flux=background_flux)
+    elseif norm == "abs"
+        pds = normalize_abs(unnorm_power, dt, n_bin)
+    elseif norm == "none"
+        pds = unnorm_power
+    else 
+        throw(ArgumentError("Incorrect Arguments"))
```
giordano, line 63: Again, more descriptive error message and test this code path.

src/fourier.jl
```diff
@@ -0,0 +1,781 @@
...
+    end
+
+    if power_type == "all"
+        return pds
+    elseif power_type == "real"
+        return real(pds)
+    elseif power_type in ["abs", "absolute"]
+        return abs.(pds)
+    else 
+        throw(ArgumentError("Incorrect Arguments"))
```
giordano, line 73: Same as above.

src/fourier.jl
```diff
@@ -0,0 +1,781 @@
...
+
+function bias_term(power1::T, power2::T, power1_noise::Real, 
+                   power2_noise::Real, n_ave::Integer;
+                   intrinsic_coherence::Real=1.0) where {T<:Union{AbstractVector{<:Real},Real}}
+    
+    if n_ave > 500
+        return 0.0 * power1
+    end
+    return @. power1 * power2 - intrinsic_coherence * (power1 - power1_noise) * (power2 - power2_noise) / n_ave
+end
```
giordano, line 86: Sounds like this should be a function of only `power1::Real, power2::Real` and `bias_term` should be called broadcasted:
```suggestion
function bias_term(power1::Real, power2::Real, power1_noise::Real, 
                   power2_noise::Real, n_ave::Integer;
                   intrinsic_coherence::Real=1.0)
    
    if n_ave > 500
        return zero(power1)
    end
    return power1 * power2 - intrinsic_coherence * (power1 - power1_noise) * (power2 - power2_noise) / n_ave
end
```

src/fourier.jl
```diff
@@ -0,0 +1,781 @@
...
+    return num / den
+end
+
+raw_coherence(cross_power::T1, power1::T2, power2::T2, power1_noise::Real, 
+              power2_noise::Real, n_ave::Integer;
+              intrinsic_coherence::Real=1.0) where 
+              {T1<:Union{AbstractVector{<:Number},Number},T2<:Union{AbstractVector{<:Real},Real}} = 
+                    _raw_coherence_single.(cross_power, power1, power2, 
+                                       power1_noise, power2_noise, n_ave;
+                                       intrinsic_coherence)
```
giordano, line 108: Again, it sounds like this should be defined on scalars, and use broadcasting on the call site

src/fourier.jl
```diff
@@ -0,0 +1,781 @@
...
+    end
+    return new_coherence                                       
+end
+
+estimate_intrinsic_coherence(cross_power:: AbstractVector{<:Complex}, 
+                             power1:: AbstractVector{<:Real}, power2:: AbstractVector{<:Real}, 
+                             power1_noise::Real, power2_noise::Real, n_ave::Integer)= 
+                                _estimate_intrinsic_coherence_single.(
+                                    cross_power, power1, power2, 
+                                    power1_noise, power2_noise, n_ave)
```
giordano, line 136: Is this `estimate_intrinsic_coherence` only calling `_estimate_intrinsic_coherence_single` broadcasted?

src/fourier.jl
```diff
@@ -0,0 +1,781 @@
...
+
+    return dRe, dIm, dphi, dG
+end
+
+function cross_to_covariance(cross_power:: AbstractVector{<:Complex}, 
+                             ref_power:: AbstractVector{<:Real}, 
+                             ref_power_noise::Real, 
+                             delta_nu::T) where {T<:Union{AbstractVector{<:Real},Real}}
+    return cross_power .* sqrt.(delta_nu ./ (ref_power .- ref_power_noise))
+end
```
giordano, line 177: Once again, the pattern in Julia is to define a function `f(x::Real) = ...` only on scalars, and then call it `f.(V)` where `V::AbstractVector{<:Real}`.  There is no need to define the function on scalars _and_ vectors, it's a useless and confusing duplicate.

src/fourier.jl
```diff
@@ -0,0 +1,781 @@
...
+
+function avg_pds_from_events(times:: AbstractVector{<:Real}, gti::AbstractMatrix{<:Real}, 
+                             segment_size::Real, dt::Real; norm::String="frac",
+                             use_common_mean::Bool=true, silent::Bool=false, 
+                             fluxes=nothing, errors=nothing)
+    if isnothing(segment_size)
+        segment_size = max(gti) - min(gti)
+    end
+    n_bin = Int64(round(segment_size / dt))
```
giordano, line 712:
```suggestion
    n_bin = segment_size ÷ dt
```

src/fourier.jl
```diff
@@ -0,0 +1,781 @@
...
+function avg_cs_from_events(times1:: AbstractVector{<:Real}, times2:: AbstractVector{<:Real}, 
+                            gti::AbstractMatrix{<:Real}, segment_size::Real, dt::Real; 
+                            norm::String="frac", use_common_mean::Bool=true, 
+                            fullspec::Bool=false, silent::Bool=false,
+                            power_type::String="all", fluxes1=nothing, fluxes2=nothing,
+                            errors1=nothing, errors2=nothing, return_auxil=false)
+    if isnothing(segment_size) 
+        segment_size = max(gti) - min(gti)
+    end
+    n_bin = Int64(round(segment_size / dt))
```
giordano, line 737:
```suggestion
    n_bin = segment_size ÷ dt
```

src/fourier.jl
Aman-Pandey-afk: Well Integer Division always rounds down right?

src/fourier.jl
Aman-Pandey-afk: Ok but I want first two arguments to always be array, should I use Ref there?

src/fourier.jl
Aman-Pandey-afk: This causes type instability, zero passes the same type as input so for Int it can return both int and float depending on condition

test/test_fourier.jl
Aman-Pandey-afk: Division by scalar doesn't need broadcasting right?

src/gti.jl
Aman-Pandey-afk: Both dt and epsilon_times_dt are float, so I guess no need to dot broadcast

src/fourier.jl
Aman-Pandey-afk: Yeah, I think I should define a single function only and call it broadcasted everywhere, the thing was API is meant to be used with vector inputs and not just broadcasted. We can mention in the docstrings though.

src/fourier.jl
giordano: Good point, then the first suggestion in https://github.com/matteobachetti/Stingray.jl/pull/2#discussion_r917182140 stands.

test/test_fourier.jl
giordano: Technically it does not, but here you're creating temporary intermediate arrays: if you broadcast everything you create a single array.  Please read https://julialang.org/blog/2017/01/moredots/

src/gti.jl
giordano: Same as https://github.com/matteobachetti/Stingray.jl/pull/2#discussion_r917256118

src/utils.jl
```diff
@@ -0,0 +1,10 @@
+import Base
+
+function sum_if_not_none_or_initialize(A,B)
+    if isnothing(A) 
+        return deepcopy((B))
+    end
+    return A + B
+end
+
+sqrt(x::Real) = x < 0.0 ? NaN : Base.sqrt(x)
```
giordano, line 10: I previously wrote a comment (in the wrong page, so I deleted it) that this was overriding a `Base` method.  This wasn't true, you're instead shadowing it (less bad, but still a bit).  What's the use case for redefining `sqrt` in this way?  If the argument is negative, an error is usually the right thing to do, instead of poising everything with `NaN`, which is hard to debug.

Aman-Pandey-afk: Well I asked Matteo on it, [Slack Message](https://stingraysoftware.slack.com/archives/C03GEPZ9ZB7/p1657274550265329?thread_ts=1657272098.400529&cid=C03GEPZ9ZB7)

Aman-Pandey-afk:
Methods like Keepat and broadcast are surely great in reducing allocations, as I can observe them with tests. I have added all the suggestions plus some changes influenced by this discussion. But, somehow the time taken is more than that before these 3 commits. (As per the ss on slack, the allocations were around 2.3 million, now 2.5 with slight increase in GC%), although opposite should happen. I think a test on other OS will provide good measure. (I'm sure this code is more efficient).

src/gti.jl
```diff
@@ -0,0 +1,98 @@
...
+
+        newbins = calculate_segment_bin_start(
+            startbin, stopbin, nbin, fraction_step=fraction_step)
+        
+        append!(spectrum_start_bins,newbins)
+    end 
+    return spectrum_start_bins, spectrum_start_bins.+nbin 
+end
+
+@resumable function generate_indices_of_segment_boundaries_unbinned(times::AbstractVector{<:Real},
```
Aman-Pandey-afk, line 73: This function is showing not to be covered by tests, but it is being called on line 177 in the `fourier.jl` file which is showing to be covered. I have tried logging some statements in this function during test runs and it is actually being executed (otherwise the tests won't pass also).

giordano: https://github.com/BenLauwens/ResumableFunctions.jl/issues/46

src/fourier.jl
```diff
@@ -0,0 +1,773 @@
...
+    end
+
+end
+
+function bias_term(power1::Real, power2::Real, power1_noise::Real, 
+                   power2_noise::Real, n_ave::Integer;
+                   intrinsic_coherence::Real=1.0)
+    
+    if n_ave > 500
+        return 0.0 * power1
```
giordano, line 83:
```suggestion
        return float(zero(power1))
```

src/fourier.jl
giordano: Now this is working correctly!

src/fourier.jl
```diff
@@ -0,0 +1,773 @@
...
+            @yield nothing
+            continue
+        end
+        if !binned
+            event_times = times[idx0:idx1-1]
+            # astype here serves to avoid integer rounding issues in Windows,
+            # where long is a 32-bit integer.
+            cts = fit(Histogram,Float64.(event_times .- s);nbins=n_bin).weights
+        else
+            cts = Float64.(fluxes[idx0+1:idx1])
```
giordano, line 221:
```suggestion
            event_times = @view times[idx0:idx1-1]
            # astype here serves to avoid integer rounding issues in Windows,
            # where long is a 32-bit integer.
            cts = fit(Histogram, float.(event_times .- s); nbins=n_bin).weights
        else
            cts = float.(@view fluxes[idx0+1:idx1])
```

src/fourier.jl
```diff
@@ -0,0 +1,773 @@
...
+            if !isnothing(errors)
+                cts = cts, errors[idx0+1:idx1]
```
giordano, line 223: Is this correct?  `cts` is changing in a hot loop from a vector to a tuple of vectors?

src/fourier.jl
```diff
@@ -0,0 +1,773 @@
...
+        end
+
+        # In the first loop, define the frequency and the freq. interval > 0
+        if isnothing(cross)
+            fgt0 = positive_fft_bins(n_bin)
+            freq = fftfreq(n_bin, dt)[fgt0]
+        end
+
+        # No need for the negative frequencies
+        unnorm_power = keepat!(unnorm_power,fgt0)
```
giordano, line 285: Am I missing something or `fgt0` here can be also `nothing`?

src/fourier.jl
```diff
@@ -0,0 +1,773 @@
...
+        ft2 = fft(flux2)
+
+        # Calculate the sum of each light curve, to calculate the mean
+        n_ph1 = sum(flux1)
+        n_ph2 = sum(flux2)
+        n_ph = sqrt(n_ph1 * n_ph2)
+
+        # Calculate the unnormalized cross spectrum
+        unnorm_power = ft1 .* conj.(ft2)
+        unnorm_pd1 = unnorm_pd2 = 0
```
giordano, line 523: Below `unnorm_pd1` and `unnorm_pd2` can become vectors, no?

src/fourier.jl
```diff
@@ -0,0 +1,773 @@
...
+        sum_of_photons1 += n_ph1
+        sum_of_photons2 += n_ph2
+
+        # Take only positive frequencies unless the user wants the full
+        # spectrum
+        if !(fullspec)
+            unnorm_power = keepat!(unnorm_power,fgt0)
+            if return_auxil
+                unnorm_pd1 = keepat!(unnorm_pd1,fgt0)
+                unnorm_pd2 = keepat!(unnorm_pd2,fgt0)
```
giordano, line 541: `keepat!` modifies the vector in-place, no need to reassign the output to itself.
```suggestion
            keepat!(unnorm_power,fgt0)
            if return_auxil
                keepat!(unnorm_pd1,fgt0)
                keepat!(unnorm_pd2,fgt0)
```

src/fourier.jl
```diff
@@ -0,0 +1,773 @@
...
+        common_variance2 /= n_ave
+        common_variance = sqrt(common_variance1 * common_variance2)
+    end
+
+    # Transform the sums into averages
+    cross /= n_ave
+    unnorm_cross /= n_ave
+    if return_auxil
+        unnorm_pds1 /= n_ave
+        unnorm_pds2 /= n_ave
```
giordano, line 620: With
```suggestion
        unnorm_pds1 ./= n_ave
        unnorm_pds2 ./= n_ave
```
you modify `unnorm_pds1` and `unnorm_pds2` in-place without allocating temporary arrays and assigning them to these variables.

src/fourier.jl
Aman-Pandey-afk: Yeah, this can be both tuple or a simple vector which is element of the `flux_iterable`. If this is costly we may separate errors into another array and `isTuple` check then becomes an `isnothing` check.

src/fourier.jl
Aman-Pandey-afk: Cross is initialized with `nothing` which means `fgt0 = positive_fft_bins(n_bin)` will always be executed first time.

src/fourier.jl
Aman-Pandey-afk: Yeah I think it is a initialization problem, these are only used when return_auxil is true and as variables have a scope outside the if block too in Julia, I might remove these initialization.

src/fourier.jl
Aman-Pandey-afk: So any place I work with operation on both scalar and vectors, I should use dot broadcast right? (Shouldn't be this then the natural behaviour)

src/fourier.jl
```diff
@@ -0,0 +1,773 @@
...
+
+    for (s, e, idx0, idx1) in fun(times, gti, segment_size)
+        if idx1 - idx0 < 2
+            @yield nothing
+            continue
+        end
+        if !binned
+            event_times = times[idx0:idx1-1]
+            # astype here serves to avoid integer rounding issues in Windows,
+            # where long is a 32-bit integer.
```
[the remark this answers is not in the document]
Aman-Pandey-afk: This was just a np.astype comment. I had kept them to understand the codes. Will change/remove them.

src/fourier.jl
Aman-Pandey-afk: Actually I should just return 0.0 now as the function is meant to be used with dot broadcast for vectors.

src/fourier.jl
matteobachetti: Indeed

src/fourier.jl
```diff
@@ -0,0 +1,771 @@
...
+    elseif norm == "frac"
+        return 2.0 / (meanrate - backrate)^2 * meanrate
+    elseif norm == "leahy"
+        return 2.0
+    elseif norm == "none"
+        return float(n_ph)
+    else 
+        throw(ArgumentError("Unknown value for norm: $norm"))
+    end
+end
```
giordano, line 21: I think a better way to represent this is to define a "normalisation" type.  Something along these lines (also with better names than `Abs` and `Frac`, I'm just copying the short names used here):
```julia
abstract type Normalization end

struct Abs{T} <: Normalization
    meanrate::T
end
poisson_level(n::Abs) = 2 * n.meanrate

struct Frac{T} <: Normalization
    meanrate::T
    backrate::T
end
poisson_level(n::Frac) = 2 / (n.meanrate - n.backrate) ^ 2 * n.meanrate

struct Leahy <: Normalization end
poisson_level(::Leahy) = 2

struct Unnormalized{T} <: Normalization
    n_ph::T
end
poisson_level(n::Unnormalized) = n_ph
```
It's idiomatic, it makes natural use of multiple dispatch (instead of dispatching on strings....), it's optimisable (the current implementation of `poisson_level` can't possibly be optimised by the compiler, since all arguments are basically `Any`), no need to abuse `nothing`, etc.

Similar things can be done elsewhere throughout the code, to make it more idiomatic and actually use Julia features, instead of writing Python code with Julia syntax.  But let's not focus on this now, this is mainly an idea for the future, to give a concrete suggestion for how to make the code more idiomatic.

giordano (approved):
I think main issues have been addressed.  However I believe the code could be much more idiomatic (I gave a concrete example for `poisson_level`), but that's for another time, although I'd like to see more code like that in the future.

src/fourier.jl
Aman-Pandey-afk: It removes use of nothing only when there is no `isnothing` check for the variable. Giving a default value with one of the other arguments of the function is one of the things heavily used in the fourier.py and it would have to be changed completely to remove the `nothing`s where this check is happening. Also, wouldn't it be more user convenient to pass the type as argument rather than making user create objects for the dispatch (Like how `fit(Histogram, args...)` works in StatsBase) 
```julia
julia> struct Abs end

julia> poisson_level(::Type{Abs}, meanrate::Real) = 2.0 * meanrate
poisson_level (generic function with 1 method)

julia> poisson_level(Abs,9)
18.0
```
And yeah, I understand that many sections can be worked upon to be more idiomatic, and I will keep that in mind writing future codes.

matteobachetti:
Thanks @Aman-Pandey-afk and @giordano, I learned a lot from this PR and its review.

PR #3, opened 2022-07-23
Implement GTI Reading and Handling

Aman-Pandey-afk:
GTI machinery and reading from file is implemented here, tests are passing and garbage collection is near to 0. Some points I would like to mention:

- Have used interconversion between Interval, Vectors and Matrices at many places which maybe less efficient but were required (like using `eachrow` sometimes which makes traversal in row order possible, which is inefficient as Julia matrices are stored in column-major order)
- I have once used nothing (in the `operations_on_gti` function) as it seemed convenient and efficient to me instead of assigning the `required_interval` an `IntervalSet` using the first element of `gti_list`.
- I removed type instabilities in most of the functions, but one seems to persist -: `setdiff(IntervalSet1, IntervalSet2)`. I looked ways to remove it but can't, it outputs an `IntervalSet` of `AbstractIntervals`.
- The `throw(ArgumentError)` is showing Any Type
- Also, in some of the functions `mapreduce` can give error if empty vectors are passed. Should I create appropriate Error throws for them? 
- The PR is based on https://github.com/matteobachetti/Stingray.jl/pull/2 . So it should be merged before this one if there are no problems except the refactoring.


src/utils.jl
```diff
@@ -4,3 +4,26 @@ function sum_if_not_none_or_initialize(A,B)
     end
     return A + B
 end
+
+function contiguous_regions(condition::AbstractVector{Bool})
+    # Find the indicies of changes in "condition"
+    d = diff(condition)
+    idx = findall(x->x != 0, d)
```
giordano, line 11:
```suggestion
    idx = findall(!iszero, d)
```

src/utils.jl
```diff
@@ -4,3 +4,26 @@ function sum_if_not_none_or_initialize(A,B)
...
+        idx = append!([1],idx)
+    end
+
+    if condition[end]
+        # If the end of condition is True, append the length of the array
+        idx = append!(idx,[condition.size+1]) # Edit
+    end
+
+    # Reshape the result into two columns
+    return reshape(idx,2,length(idx) ÷ 2)'
```
giordano, line 28: What happens if the length of `idx` isn't even?

src/gti.jl
```diff
@@ -1,3 +1,194 @@
...
+    safe_interval::AbstractVector{<:Real}=[0,0], dt::AbstractVector{<:Real}=Float64[])
+    
+    if length(time) != length(condition)
+        throw(ArgumentError("The length of the condition and 
+        time arrays must be the same."))
+    end
+
+    idxs = contiguous_regions(condition)
+
+    if(isempty(dt))
```
giordano, line 112: No need to use parentheses around conditionals
```suggestion
    if isempty(dt)
```

src/gti.jl
```diff
@@ -1,3 +1,194 @@
...
+
+    return mask, mapreduce(permutedims, vcat, keepat!(new_gtis,new_gti_mask))
+end
+
+function create_gti_from_condition(time::AbstractVector{<:Real}, condition::AbstractVector{Bool};
+    safe_interval::AbstractVector{<:Real}=[0,0], dt::AbstractVector{<:Real}=Float64[])
+    
+    if length(time) != length(condition)
+        throw(ArgumentError("The length of the condition and 
+        time arrays must be the same."))
```
giordano, line 107: This is printed as
```
julia> throw(ArgumentError("The length of the condition and 
               time arrays must be the same."))
ERROR: ArgumentError: The length of the condition and 
        time arrays must be the same.
```
which is quite odd.  Either don't break the line, or don't add indentation to the second line.

src/gti.jl
```diff
@@ -1,3 +1,194 @@
...
+            "This GTI has overlaps"
+        ))
+    end
+end
+
+function create_gti_mask(times::AbstractVector{<:Real},gtis::AbstractMatrix{<:Real};
+                         safe_interval::AbstractVector{<:Real}=[0,0], min_length::Real=0,
+                         dt::Real = -1, epsilon::Real = 0.001)
+
+    if length(times) == 0
```
giordano, line 58:
```suggestion
    if isempty(times)
```
should be the same, right?

src/gti.jl
```diff
@@ -1,3 +1,194 @@
...
+
+    for interval in required_interval.items
+        push!(final_gti, [first(interval), last(interval)])
+    end
+
+    return mapreduce(permutedims, vcat, final_gti)
+end
+
+function get_btis(gtis::AbstractMatrix{<:Real})
+    if length(gtis) == 0
```
giordano, line 160:
```suggestion
    if isempty(gtis)
```

src/gti.jl
```diff
@@ -1,3 +1,194 @@
...
+        throw(ArgumentError("Empty GTI and no valid start_time and stop_time"))
+    end
+    return get_btis(gtis, gtis[1,1], gtis[end,2])
+end
+
+function get_btis(gtis::AbstractMatrix{T}, start_time, stop_time) where {T<:Real}
+    if length(gtis) == 0
```
giordano, line 167:
```suggestion
    if isempty(gtis)
```

src/gti.jl
```diff
@@ -1,3 +1,194 @@
...
+        return T[start_time stop_time]
```
giordano, line 168: Did you return a matrix instead of a vector on purpose?

src/gti.jl
```diff
@@ -1,3 +1,194 @@
+function get_total_gti_length(gti::AbstractMatrix{<:Real}; minlen::Real=0.0)
+    lengths = diff(gti; dims =2)
+    return sum(x->x > minlen ? x : zero(x), lengths)
+end
+
+function load_gtis(fits_file::String, gtistring::String="GTI")
+    lchdulist = FITS(fits_file)
+    gtihdu = lchdulist[gtistring]
+    gti = get_gti_from_hdu(gtihdu)
+    close(lchdulist)
```
giordano, line 10: I think you can do
```suggestion
    gti = FITS(fits_file) do lchdulist
        gtihdu = lchdulist[gtistring]
        get_gti_from_hdu(gtihdu)
    end
```

src/utils.jl
Aman-Pandey-afk: Length of idx will always be even as if there is 010 or 101 in the body (excluding the ends) it diff is nonzero at two places. Else it start or end will be 1 and thus they are added .

PR #7, opened 2022-08-29
Implement EventList

Aman-Pandey-afk:
This is a basic implementation of `EventList` with essential APIs and tests
Things remaining:
- Implement EventList Writing
- LightCurve Methods
- Plot Testing
- Optimizing/Refactoring the Code

Cross-spectra/Periodograms APIs and documentation will be implemented soon in next PRs

### DIFF
--- a/src/events.jl
+++ b/src/events.jl
@@ -1,0 +1,136 @@
+@with_kw mutable struct EventList{T1<:AbstractVector,T2<:AbstractVector,T3<:AbstractMatrix}
+    time::T1
+    energy::T2=Float64[]
+    ncounts::Int=0
+    mjdref::Float64=0
+    dt::Float64=0
+    notes::String=""
+    gti::T3=reshape(Float64[],0,2)
+    PI::Vector{Int64}=Int[]
+    high_precision::Bool=false
+    mission::String=""
+    instr::String=""
+    header::String=""
+    detector_id::Vector{String}=String[]
+    ephem::String=""
+    timeref::String=""
+    timesys::String=""
+end
+
+function read(filename::String, format::String)
+    if format=="fits"
+        return load_events_from_fits(filename)
+    else
+        throw(ArgumentError("File not supported still."))
+    end
+end
+
+function write(ev::EventList, filename::String, format::String)
+    if format=="fits"
+        write_events_to_fits(filename, ev)
+    else
+        throw(ArgumentError("File not supported still."))
+    end
+end
+
+function to_lc(ev::EventList, dt)
+    tstart = ev.gti[1][1]
+    tseg = ev.gti[end][end]
+    return to_lc(ev, dt, tstart, tseg)
+end
+
+to_lc(ev::EventList, dt::Real,
+      tstart::Real, tseg::Real) = make_lightcurves(ev.time, dt, tstart=tstart,
+                                                   gti=ev.gti, tseg=tseg,
+                                                   mjdref=ev.mjdref)
+
+function join(ev1::EventList, ev2::EventList)
+    new_ev = EventList(time=vcat(ev1.time, ev2.time))
+    if ev1.dt!=ev2.dt
+        new_ev.dt = max(ev1.dt,ev2.dt);
+    end
+
+    # Tolerance for MJDREF:1 microsecond
+    if !isapprox(ev1.mjdref, ev2.mjdref, atol=1e-6 / 86400)
+        ev2.mjdref = ev1.mjdref
+    end
+
+    order = sortperm(new_ev.time)
+    new_ev.time = new_ev.time[order]
+
+    if !isempty(ev1.PI) || !isempty(ev2.PI)
+        if isempty(ev1.PI)
+            ev1.PI = zero(ev1.time)
+        end
+        if isempty(ev2.PI)
+            ev2.PI = zero(ev2.time)
+        end
+        new_ev.PI = vcat(ev1.PI, ev2.PI)
+        new_ev.PI = new_ev.PI[order]
+    else
+        new_ev.PI = Float64[]
+    end
+
+    if !isempty(ev1.energy) || !isempty(ev2.energy)
+        if isempty(ev1.energy)
+            ev1.energy = zero(ev1.time)
+        end
+        if isempty(ev2.energy)
+            ev2.energy = zero(ev2.time)
+        end
+        new_ev.energy = vcat(ev1.energy, ev2.energy)
+        new_ev.energy = new_ev.energy[order]
+    else
+        new_ev.energy = Float64[]
+    end
+
+    if !isempty(ev1.gti) || !isempty(ev2.gti)
+        if isempty(ev1.gti)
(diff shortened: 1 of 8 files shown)


Aman-Pandey-afk:
These were minor mistakes (like not uncommenting some codes, mostly 1-2 lines) so force pushed in the same commit.

src/events.jl
```diff
@@ -0,0 +1,136 @@
+@with_kw mutable struct EventList{T1<:AbstractVector,T2<:AbstractVector,T3<:AbstractMatrix}
+    time::T1
+    energy::T2=Float64[]
+    ncounts::Int=0
+    mjdref::Float64=0
+    dt::Float64=0
+    notes::String=""
+    gti::T3=reshape(Float64[],0,2)
+    PI::Vector{Int64}=Int[]
```
giordano, line 9: `Int` isn't necessarily `Int64` (it is on 64-bit systems, but not on 32-bit ones).  You either use `Int` or `Int64` everywhere, you have to be consistent.  If that helps in the choice, literal integers are of type `Int` (so they're actual type depends on the system where the code is run)

src/events.jl
```diff
@@ -0,0 +1,136 @@
...
+    ephem::String=""
+    timeref::String=""
+    timesys::String=""
+end
+
+function read(filename::String, format::String)
+    if format=="fits"
+        return load_events_from_fits(filename)
+    else
+        throw(ArgumentError("File not supported still."))
```
giordano, line 24: I'd make the error message more explicit:
```suggestion
        throw(ArgumentError("File format $(format) not yet supported."))
```
Same below.

src/events.jl
```diff
@@ -0,0 +1,136 @@
...
+function write(ev::EventList, filename::String, format::String)
+    if format=="fits"
+        write_events_to_fits(filename, ev)
+    else
+        throw(ArgumentError("File not supported still."))
+    end
+end
+
+function to_lc(ev::EventList, dt)
+    tstart = ev.gti[1][1]
```
giordano, line 37:
```suggestion
    tstart = ev.gti[begin][begin]
```

src/events.jl
```diff
@@ -0,0 +1,136 @@
...
+        if isempty(ev1.gti)
+            ev1.gti = [ev1.time[begin] - ev1.dt/2 ev1.time[end] + ev1.dt/2]
+        end
+        if isempty(ev2.gti)
+            ev2.gti = [ev2.time[begin] - ev2.dt/2 ev2.time[end] + ev2.dt/2]
+        end
+        new_ev.gti = operations_on_gtis([ev1.gti,ev2.gti], intersect)
+        if isempty(new_ev.gti)
+            @warn "GTIs in these two event lists do not overlap at all.
+            Merging instead of returning an overlap"
```
giordano, line 97: This is printed like this:
```
julia> @warn "GTIs in these two event lists do not overlap at all.
                   Merging instead of returning an overlap"
┌ Warning: GTIs in these two event lists do not overlap at all.
│             Merging instead of returning an overlap
└ @ Main REPL[55]:1
```
which shows a quite odd indentation.  Use a [triple-quoted string](https://docs.julialang.org/en/v1/manual/strings/#Triple-Quoted-String-Literals) instead
```suggestion
            @warn """
            GTIs in these two event lists do not overlap at all.
            Merging instead of returning an overlap.
            """
```

src/events.jl
```diff
@@ -0,0 +1,136 @@
...
+            new_ev.gti = operations_on_gtis([ev1.gti,ev2.gti], union)
+        end
+    else
+        new_ev.gti = reshape([],0,2)
```
giordano, line 101: `[]` is of type `Any`, is that desired?

src/events.jl
```diff
@@ -0,0 +1,136 @@
...
+    end
+
+    for attr in [:mission, :instr]
```
giordano, line 104: For such small collections use tuples, which don't allocate memory, instead of vectors
```suggestion
    for attr in (:mission, :instr)
```

src/events.jl
```diff
@@ -0,0 +1,136 @@
...
+end
+
+function apply_mask(ev::EventList, mask::AbstractVector{T};) where {T<:Union{Bool,Int}}
+    for field in fieldnames(EventList)
+        fval = getfield(ev, field)
+        if fval isa AbstractVector && !isempty(fval) 
+            setfield!(ev, field, deepcopy(fval)[mask])
+        end
+    end
+end
```
giordano, line 136: Missing newline
```suggestion
end

```

src/io.jl
```diff
@@ -0,0 +1,67 @@
...
+        ev
+    end
+    return a
+end
+
+function write_events_to_fits(filename::String, ev::EventList)
+    FITS(filename,"w") do hduList
+        
+        tkeys = String[]
+        values = []
```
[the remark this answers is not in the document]
Aman-Pandey-afk: They are values of keys of `EventLists` to be put in HDU header. They can be different

src/utils.jl
```diff
@@ -27,3 +27,90 @@ function contiguous_regions(condition::AbstractVector{Bool})
...
     return reshape(idx,2,length(idx) ÷ 2)'
 end
+
+function rebin_data(x::AbstractVector{<:Real}, y::AbstractVector{<:Integer}, dx_new::Real;
+                    y_err::AbstractVector{<:Real}=Float64[], method::String = "sum", dx::Real=0)
+    if isempty(y_err)
+        y_err = zero(y)
+    end
+
+    if dx isa AbstractVector
```
giordano, line 37: How can `dx` be an `AbstractVector` if it's annotated as `Real`?

src/utils.jl
```diff
@@ -27,3 +27,90 @@ function contiguous_regions(condition::AbstractVector{Bool})
...
+            post_frac = (xmax-xedges[max_ind-1])/dx_post
+            output[i] += y[max_ind-1]*post_frac
+            outputerr[i] += y_err[max_ind-1]*post_frac
+            step_size[i] += post_frac
+        end
+    end
+
+    if method in ["mean", "avg", "average", "arithmetic mean"]
+        ybin = output / step_size
+        ybinerr = sqrt.(outputerr) / step_size
```
giordano, line 91: Fuse dot operations
```suggestion
        ybinerr = sqrt.(outputerr) /. step_size
```

src/utils.jl
```diff
@@ -27,3 +27,90 @@ function contiguous_regions(condition::AbstractVector{Bool})
...
+    end
+
+    dx_var = Statistics.var(dx_old) / Statistics.mean(dx_old)
+
+    if size(dx_old) == 1 || dx_var < 1e-6
+        new_step = step_size[1]
+    end
+
+    new_x0 = (x[1] - (0.5 * dx_old[1])) + (0.5 * dx_new)
+    xbin = 1:length(ybin) * dx_new .+ new_x0
```
giordano, line 113: Just double checking: are you entirely sure this is parsed as you expect?

PR #10, opened 2025-03-07
Sampledata

kashish2210:
# sampledata  
**Adding Sample_data_test in julia**  

## Description  
This pull request addresses the following issues:  
- added sample [lc] from `stingraysoftware` 
- added `test_sampledata.jl` and `sampledata.jl`

## How to Test  
1. Run the test suite using:  
 ```sh
  julia runtests.jl
```
## output:
output_sample_data_test


ankitkhushwaha:
What is size of txt file u used. Plz make sure it should be less than 100 kb or it should hosted in some where else.

See contributing guide of stingray.

PR #21, opened 2025-03-20
Simplify CI

fjebaker:
Reducing how many different OSs and Julia versions we are running the CI on for the purposes of developing the package. Julia nightly is failing on this package, but is not really a version we need to worry about supporting at this stage of the development.

This also bumps all the CI action versions.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -46,6 +46,7 @@ jobs:
       - uses: julia-actions/setup-julia@v2
         with:
           version: '1'
+      - uses: julia-actions/cache@v2
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-docdeploy@v1
         env:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -18,12 +18,9 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.7'
-          - 'nightly'
+          - '1.11'
         os:
           - ubuntu-latest
-          - macOS-latest
-          - windows-latest
         arch:
           - x64
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -45,8 +42,8 @@ jobs:
     permissions:
       contents: write
     steps:
-      - uses: actions/checkout@v3
-      - uses: julia-actions/setup-julia@v1
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v2
```
giordano, line 46: Might want to use cache here as well

fjebaker:
The CI will fail on this branch as it still has the Metadata.jl dependency.

giordano (commented):
May be good to install dependabot too

.github/workflows/CI.yml
fjebaker: Good point!

kashish2210:
i was getting this error locally to for that I have tried 
```
using Pkg
rm("Manifest.toml") 
#then entered new project.toml
Pkg.instantiate()     # Reinstall packages fresh
Pkg.resolve()       
Pkg.gc()            
Pkg.precompile()      
```
for this we can try using this in ci **LOCALLY**
```
- name: Clear Julia Package Cache
  run: |
    rm -rf ~/.julia/compiled/
    rm -rf ~/.julia/packages/
```
to clear previous cachee

kashish2210:
one additional worflow u can add 
```
lint:
    name: Julia Code Formatting Check
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: julia-actions/setup-julia@v2
      - run: julia -e 'using JuliaFormatter; format(".")'
      ```

matteobachetti (approved):
I think it's fair to eliminate the nightly dependency at this stage. We will add it back when the package is a little more stable and we care about long-term maintenance.

PR #22, opened 2025-03-21
Add Documentation for Setup and First Interaction

JawadAhmadCS:
This PR adds documentation to assist new contributors in setting up the environment and understanding the initial interaction with the project.  

> Verified all steps by completely uninstalling and reinstalling the environment.


JawadAhmadCS:
Hi @fjebaker & @matteobachetti  ,  

I have added the initial documentation and verified all installation steps by completely reinstalling the environment.  

### Observation:
- Encountered an **error with the `Stingray` package**: `ERROR: LoadError: ArgumentError: Package Stingray not found in current path.`  
  - I resolved this by manually installing `Stingray`.  

Do you think I should provide a solution for this issue as well, or is it unnecessary?  if its necessary than please seggest me where i provide it solution.

image

JawadAhmadCS:
@fjebaker Please **review** and provide your **feedback**...Thanks!

JawadAhmadCS:
I noticed that this project hasn’t been deployed . If it gets deployed, the documentation can be accessed at: 

**http://docs.stingray.science/Stingray.jl/**

kashish2210 (commented):
I told you to use a separate project.toml for this 
You may avoid this type of usage as it can disturb the structure of codebase

kashish2210:
Also the previous dependency if you are encountering this means your genral.toml is broken you can either delete your all manifest.toml locally or you can use Pkg.rm(["matadata"]) and for array interface error you may need to delete your local manifest and reinstall it 
Best of luck :)

kashish2210:
> I noticed that this project hasn’t been deployed . If it gets deployed, the documentation can be accessed at:
> 
> **http://docs.stingray.science/Stingray.jl/**

try using this
```
Pkg.add(url="https://github.com/matteobachetti/Stingray.jl")
Pkg.develop("Stingray")

JawadAhmadCS:
> I told you to use a separate project.toml for this You may avoid this type of usage as it can disturb the structure of codebase but this seems okay for running locally

Thanks for the review @kashish2210 ! Actually, you're thinking that the `project.toml` is the same for both `docs` and `Stingray.jl`, but that's not the case. They have separate files, and I didn't make any changes to `docs/project.toml`; it's the same as before.  

The changes you're seeing are in `Stingray.jl/project.toml`, and they happened because I uninstalled Julia 1.7 and installed 1.11. This caused all dependencies to be reinstalled, which led to some structural changes in `Stingray.jl`. However, the content remains the same; only the lines have shifted, so the structure isn't actually disturbed.

JawadAhmadCS:
> Also the previous dependency if you are encountering this means your genral.toml is broken you can either delete your all manifest.toml locally or you can use Pkg.rm(["matadata"]) and for array interface error you may need to delete your local manifest and reinsiate it Best of luck :)

Actually, I created this branch without updating it with the latest changes, so the new dependencies weren't updated. That's why this issue occurred.

Issue #27, opened 2025-03-21
Reading events files


fjebaker:
Event lists will be a list of times when photons arrived into the detector, and will likely also contain some information about what energy they have. 

We need a simple routine that, given an events file, reads it into memory.

It would be good to have an abstraction in the code that represents this. Maybe something like
```julia
struct EventList{T,Meta}
    times::Vector{T}
    energies::Vector{T}
    metadata::Meta
end

function readevents(path; T = Float64)
    # ... returns an EventList
end
```
This can then have metadata about the dataset (e.g. telescope, object ID, etc.) added at a later time.

fjebaker:
@kashish2210 will work on this 👍

fjebaker:
Hi @kashish2210 I'm responding to your comment from #26 on how you are planning to implement this feature.

Your plan looks good! But it's a little too general for what this needs to be. Currently we don't need something that matches all of the Python Stingray features, we just want something we can work towards our goals with. The overall aim of Stringray.jl at the moment is to see where Julia can provide features that the Python implementation can't (e.g. using AD compiled to machine code during power spectra fits). We don't need to reproduce every Python feature to estimate what ballpark the Julia performance can offer :)

So I'd encourage you to focus on just implementing the bare minimum that we need to load in an events file as a list of times and energies. My guess is that we only need a simple structure, and a single function that reads in the data into that structure. We don't need a logger and so on at the moment.

The metadata can be included later if you follow the above struct design, but I am happy to discuss or iterate on the design with you! I think at this stage it's premature to add all of the metadata fields directly into the struct. 

Let me know what you think!

kashish2210:
thank u @fjebaker for mentioning this
`The overall aim of Stingray.jl at the moment is to see where Julia can provide features that the Python implementation can't (e.g. using AD compiled to machine code during power spectra fits). `
now I know what I should focus on from now I will focus on just implementing the bare minimum functions 
**Further I need a guide about what exactly we are targeting that Python does not support** 
Thanks for reviewing my plan even though I created `loggingconfig.jl` and implemented successful tests ,I will move it to draft for now

fjebaker:
Completed in #30

PR #30, opened 2025-03-22
readevent

kashish2210:
#27 
my [workfloww](https://github.com/kashish2210/notebooks/blob/main/eventlist.ipynb)
i have tested using [kplr011446443.fits](https://archive.stsci.edu/pub/kepler/lightcurves/)
for direct getting this FITS file you can visit my [github](https://github.com/kashish2210/notebooks/blob/main/kplr011446443.fits)

further I have created 21 test cases that includes :
- to test monol_testA.evt from `test/data`

@fjebaker @JawadAhmadCS your review required :)


kashish2210:
I need a suggestion of fits that I should use that has `energy values`

kashish2210:
I was thinking of adding the case for .txt and .gz files but as per your comment I have done minimal implementation in future we can add these if needed

src/events.jl
```diff
@@ -0,0 +1,43 @@
...
+                if "TIME" in colnames
+                    times = convert(Vector{T}, read(hdu, "TIME"))
+                end
+                if "ENERGY" in colnames
+                    energies = convert(Vector{T}, read(hdu, "ENERGY"))
+                end
+            end
+        end
+    end
+
```
fjebaker, line 40: I'd suggest adding some kind of warning or even error if the `times` and `energies` vectors are still empty at this point, so the caller knows that their FITS file may have been invalid.

src/Stingray.jl
```diff
@@ -32,4 +32,7 @@ export bin_intervals_from_gtis
 
 include("utils.jl")
 
-end 
+include("events.jl")
+export readevents
```
fjebaker, line 36:
```suggestion
export EventList, Meta, readevents
```

src/events.jl
```diff
@@ -0,0 +1,43 @@
+struct Meta
```
fjebaker, line 1: I'd maybe suggest renaming this struct to something like `DictMetadata`?

fjebaker (commented):
Hi @kashish2210 this looks great! Just two little comments and a suggestion to export your structs.

fjebaker:
Also, please could you add docstrings and struct field documentation! That will help us maintain this code going forwards :+1:

kashish2210:
yup I have added the necessary docstrings and warns and updated the test cases to suppress warns while testing for advance implementations we will include `logconfig.jl`

src/events.jl
```diff
@@ -0,0 +1,90 @@
...
+        The numeric type for time and energy values. Default is Float64.
+    Returns
+    -------
+    EventList{T}
+        An EventList structure containing the extracted data.
+    Notes
+    -----
+    The function extracts TIME and ENERGY columns from any TableHDU in the FITS file.
+    All headers from each HDU are collected into the metadata field.
+    """
```
fjebaker, line 52: Julia function docstrings go above the function definition. It is also more
standard to document the parameters in prose instead of as lists, as much of
their semantics is in the function signature.

```suggestion

"""
    readevents(path; T = Float64)

Read event data from a FITS file into an EventList structure. The `path` is a
string that points to the location of the FITS file. `T` is used to specify
which numeric type to convert the data to.

Returns an [`EventList`](@ref) containing the extracted data.

## Notes

The function extracts `TIME` and `ENERGY` columns from any TableHDU in the FITS
file. All headers from each HDU are collected into the metadata field.
"""
function readevents(path; T = Float64)
```

src/events.jl
```diff
@@ -0,0 +1,90 @@
...
+    headers::Vector{Dict{String,Any}}
+end
+
+"""
+    EventList{T}
+
+A structure containing event data from a FITS file.
+
+Fields
+------
```
fjebaker, line 20:
```suggestion
## Fields

```

fjebaker (changes requested):
Thanks for adding the docstrings! I have added some suggestion for how these are usually done in Julia.

kashish2210:
> Thanks for adding the docstrings! I have added some suggestion for how these are usually done in Julia.

thanks for suggestions

src/events.jl
```diff
@@ -0,0 +1,85 @@
+"""
+    DictMetadata
+
+A structure containing metadata from FITS file headers.
+
+Fields
+------
```
fjebaker, line 7: Sorry I missed just this one:

```suggestion
## Fields

```

fjebaker (approved):
Brilliant! Nice work @kashish2210

kashish2210:
> Brilliant! Nice work @kashish2210

thanks :)

Issue #28, opened 2025-03-21
Converting events to lightcurves


fjebaker:
Builds on #27. A lightcurve is a list of time bins, event counts in that bin, the error on the counts in the bin. 

Maybe something like
```julia
struct LightCurve{T}
    timebins::Vector{T}
    counts::Vector{T}
    count_error::Vector{T}
    err_method::Symbol # or something like an enum? identifies what the method for computing the error was
end
```

The constructor should take an eventlist and return the lightcurve. We can use a lightcurve to represent the source or background, and will track the context seperate from the lightcurve.

We need to be conscious of how GTIs will work with the lightcurves. Either, we can have the lightcurve know it's GTI and use it as a mask when computing values. Or, we return a new lightcurve for each GTI, chopping up a larger lightcurve into smaller segments (we can maybe change the struct to use something less concrete than Vector so we can pass views around to be memory efficient).

I think I prefer the latter for GTIs but please discuss if you have other opinions! We'll not worry about GTIs just yet in this issue though.

fjebaker:
@JawadAhmadCS will work on this 👍

JawadAhmadCS:
Hey @fjebaker, Thanks for assigning me to #28! I accept the responsibility for implementing this feature and will ensure it is completed efficiently. Since #28 depends on #27, I’ll wait for @kashish2210 to define the eventlist structure before finalizing the implementation. Meanwhile, I’ll begin prototyping the core logic. Let me know if there are any specific considerations to keep in mind!

JawadAhmadCS:
Hi @fjebaker i have completed my this assigned task so please review

fjebaker:
Hi @JawadAhmadCS thanks! I will get round to reviewing it when I have the time -- it's on my list of things to do :)

JawadAhmadCS:
Ok no problem @fjebaker I spent a lot of time and research understanding the theory and implementing the logic. that’s why I asked for a review

kashish2210:
Let's close this to as it done #51

PR #31, opened 2025-03-22
events to lightcurve

JawadAhmadCS:
Issue #28 

# Use Case
here how this feature use after addition
🔗 Highlighted Use Case Notebook: [Notebook Link](https://github.com/JawadAhmadCS/notebooks/blob/main/lightcurve%20usecase.ipynb)

---

🔗 Julia Workflow Notebook: [Notebook Link](https://github.com/JawadAhmadCS/notebooks/blob/main/lightcurve_jl.ipynb)
🔗 Python Notebook (Stingray  lightcurve usecase for understandng): [Notebook Link](https://github.com/JawadAhmadCS/notebooks/blob/main/lightcurve_py.ipynb)

image
image
image


src/lightcurve.jl
```diff
@@ -0,0 +1,96 @@
...
+    if bin_size > (max_time - min_time) / 10
+        @warn "The bin size is relatively large compared to the time range. This may result in a low number of bins."
+    elseif bin_size < 0.001
+        @warn "The bin size is extremely small. This might lead to excessive error or unnecessary fine-grained bins."
+    end
+
+    # Check for valid error method
+    if err_method != :poisson && err_method != :sqrtN
+        @warn "Unrecognized error method. Using default ':poisson' method."
+        err_method = :poisson  # Default to :poisson if invalid
```
fjebaker, line 59:
```suggestion
        throw("Unrecognized error method: $err_method")
```
If the user passes an invalid symbol, that should be an error. You might want to raise a meaningful exception here, or just `throw` for now.

src/lightcurve.jl
```diff
@@ -0,0 +1,96 @@
...
+    err_method::Symbol
+end
+
+"""
+    create_lightcurve(eventlist::Events.EventList{T}, bin_size::T; err_method::Symbol=:poisson) where T
+Generate a light curve from an [`EventList`](@ref) by binning event times.
+## Arguments
+- `eventlist::Events.EventList{T}`: The input event list.
+- `bin_size::T`: The size of each time bin.
+- `err_method::Symbol`: The method for error estimation (`:poisson` or `:sqrtN`).
```
fjebaker, line 32: I think we should only have `:poisson` symbol here, not `:poisson` and `:sqrtN`, as they are the same thing, and it might lead to confusion to have both mean the same.

src/lightcurve.jl
```diff
@@ -0,0 +1,96 @@
...
+    if err_method != :poisson && err_method != :sqrtN
+        @warn "Unrecognized error method. Using default ':poisson' method."
+        err_method = :poisson  # Default to :poisson if invalid
+    end
+
+    # Binning events and calculating counts and errors
+    for (i, bin_start) in enumerate(bins[1:end-1])
+        bin_end = bins[i + 1]
+        events_in_bin = filter(x -> bin_start <= x < bin_end, eventlist.times)
+        counts[i] = length(events_in_bin)
```
fjebaker, line 66:
```suggestion
        counts[i] = count(x -> bin_start <= x < bin_end, eventlist.times)
```
I am pretty sure this should work, and doesn't need to allocate a new array for each iteration :+1:

src/lightcurve.jl
```diff
@@ -0,0 +1,96 @@
...
+
+        if err_method == :poisson || err_method == :sqrtN
+            errors[i] = sqrt(counts[i])
+        else
+            errors[i] = 0.0
```
fjebaker, line 71:
```suggestion
```
No need to zero since `errors` was allocated with `zeros`.

src/lightcurve.jl
```diff
@@ -0,0 +1,96 @@
...
+    # Plot data with error ribbon
+    return _plot(
+        timebins, counts,
+        ribbon=errors,
+        label="Event Counts",
+        xlabel="Time",
+        ylabel="Counts",
+        legend=:topright
+    )
+end
```
fjebaker, line 94: This is good! But this is out of scope for #28. Could you remove it from this PR, along with the `Plots` dependency?

test/test_lightcurve.jl
```diff
@@ -0,0 +1,171 @@
...
+                    @test data.filename == test_filepath
+                    @test length(data.metadata.headers) > 0
+                finally
+                    global_logger(old_logger)
+                end
+            end
+        else
+            @info "Test file '$(test_filepath)' not found. Skipping this test."
+        end
+    end
```
fjebaker, line 61: Many of these tests are already in the `test_events.jl` file. You can either add your tests directly to `test_events.jl` to avoid duplicating, or remove the bits that are not relevant to testing your lightcurve implementation.

test/data/sample_events.fits
fjebaker, line 1: Where did this file come from? I also can't see where it is being used. Can it be removed?

fjebaker (changes requested):
Hi @JawadAhmadCS ! This is a great start! I've left a few comments about cleaning up parts of your implementation to make it more Julia-idiomatic, and about handling your tests.

src/lightcurve.jl
```diff
@@ -0,0 +1,96 @@
...
+import ..Events  # Import EventList from events.jl
+using Plots  # Importing the Plots library for graph plotting
+
+export LightCurve, create_lightcurve, plot_lightcurve
+
+"""
+    LightCurve{T}
+A structure representing a light curve, which is a time series of event counts.
+## Fields
+- `timebins::Vector{T}`: Time bins for the light curve.
```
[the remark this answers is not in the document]
JawadAhmadCS: Yes, now it's correctly formatted according to the Markdown spec.

src/lightcurve.jl
```diff
@@ -0,0 +1,96 @@
...
+"""
+    create_lightcurve(eventlist::Events.EventList{T}, bin_size::T; err_method::Symbol=:poisson) where T
+Generate a light curve from an [`EventList`](@ref) by binning event times.
+## Arguments
+- `eventlist::Events.EventList{T}`: The input event list.
+- `bin_size::T`: The size of each time bin.
+- `err_method::Symbol`: The method for error estimation (`:poisson` or `:sqrtN`).
+## Returns
+A [`LightCurve`](@ref) instance containing binned event counts.
+## Notes
```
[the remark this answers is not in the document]
JawadAhmadCS: done!

src/lightcurve.jl
JawadAhmadCS: I added both poisson and sqrtN to provide flexibility but I understand they are equivalent .now i  remove it to avoid confusion

src/lightcurve.jl
```diff
@@ -0,0 +1,96 @@
...
+
+    counts = zeros(T, n_bins)
+    errors = zeros(T, n_bins)
+
+    # Warning if bin size is too large or too small
+    if bin_size > (max_time - min_time) / 10
+        @warn "The bin size is relatively large compared to the time range. This may result in a low number of bins."
+    elseif bin_size < 0.001
+        @warn "The bin size is extremely small. This might lead to excessive error or unnecessary fine-grained bins."
+    end
```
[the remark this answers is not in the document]
JawadAhmadCS: checks for large or small bin sizes removed as requested. we now trust the caller to handle this

src/lightcurve.jl
JawadAhmadCS: yes you`re right its creating redundantancy..but now i have removed it so its done now

src/lightcurve.jl
JawadAhmadCS: Apologies for this. Actually, while I was developing the logic for lightcurve in Jupyter, I was also plotting graphs. thats why i also added this  here

test/test_lightcurve.jl
JawadAhmadCS: apologies for the confusion. during my testing, I was encountering some errors with the event tests due to some changes. to troubleshot i added these tests to the lightcurve tests section. however I forgot to remove them afterward.

test/data/sample_events.fits
JawadAhmadCS: When I completed the full implementation for the lightcurve, I create the test sample_file for checking the use case of my developed feature in jupyter. At that time, when I ran the Jupyter notebook command, i was in this directory so its created here. I have now deleted it as it's no longer needed.

src/lightcurve.jl
JawadAhmadCS: done!

src/events.jl
```diff
@@ -1,3 +1,8 @@
+module Events
+export EventList, readevents
```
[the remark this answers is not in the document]
JawadAhmadCS: When I was working on implementing the graph plotting module diirectly, I ran into some errors. As a temporary fix, I wrapped the code in a module. later everything was working so I didn't remove it. I'll now go ahead and remove the modules

src/lightcurve.jl
JawadAhmadCS: Yes 💯 now its updted according seggetison

test/test_lightcurve.jl
```diff
@@ -0,0 +1,35 @@
...
+    # Test reading the sample file
+    eventlist = readevents(sample_file)  # No module prefix since it's directly available
+    @test eventlist.filename == sample_file
+    @test length(eventlist.times) == length(times)
+    @test length(eventlist.energies) == length(energies)
+
+    # Create a LightCurve
+    bin_size = 1.0
+    err_method = :poisson
+    lightcurve = create_lightcurve(eventlist, bin_size, err_method=err_method)
```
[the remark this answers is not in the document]
JawadAhmadCS: done

src/lightcurve.jl
```diff
@@ -0,0 +1,67 @@
...
+    if err_method != :poisson && err_method != :sqrtN
+        throw("Unrecognized error method: $err_method")  # Throw error if the method is unrecognized
+    end
+
+    # Binning events and calculating counts and errors
+    for (i, bin_start) in enumerate(bins[1:end-1])
+        bin_end = bins[i + 1]
+        counts[i] = count(x -> bin_start <= x < bin_end, eventlist.times)
+
+        if err_method == :poisson || err_method == :sqrtN
```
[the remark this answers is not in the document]
JawadAhmadCS: done

src/lightcurve.jl
```diff
@@ -0,0 +1,67 @@
...
+    max_time = maximum(eventlist.times)
+
+    bins = min_time:bin_size:max_time
+    n_bins = length(bins) - 1
+
+    counts = zeros(T, n_bins)
+    errors = zeros(T, n_bins)
+
+    # Check for valid error method
+    if err_method != :poisson && err_method != :sqrtN
```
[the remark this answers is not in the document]
JawadAhmadCS: done

test/test_lightcurve.jl
```diff
@@ -0,0 +1,34 @@
...
+
+    # Test different bin sizes
+    for bin_size in [1.0, 0.5, 2.0, 1.3]
+        lightcurve = create_lightcurve(eventlist, bin_size, err_method=:poisson)
+
+        # Validate LightCurve properties
+        @test all(lightcurve.counts .>= 0)
+        @test lightcurve.err_method == :poisson
+        @test length(lightcurve.timebins) == length(lightcurve.counts)
+    end
```
fjebaker, line 33: This is fine as a smoke test, but ideally this test should verify that it has indeed calculated the correct number of counts per bin for the different `bin_size`, i.e. as a regression test. 

Could you modify this so that it explicitly tests that it has the correct number of counts in each bin, for each `bin_size`?

fjebaker (changes requested):

src/lightcurve.jl
```diff
@@ -0,0 +1,67 @@
...
+    counts = zeros(T, n_bins)
+    errors = zeros(T, n_bins)
+
+    # Check for valid error method
+    if err_method != :poisson
+        throw("Unrecognized error method: $err_method")  # Throw error if the method is unrecognized
+    end
+
+    # Binning events and calculating counts and errors
+    for (i, bin_start) in enumerate(bins[1:end-1])
```
matteobachetti, line 56: Is there an efficient histogram implementation in Julia? I guess it would be good to use some known fast processing rather than reinventing it.

src/lightcurve.jl
```diff
@@ -0,0 +1,67 @@
...
+        bin_end = bins[i + 1]
+        counts[i] = count(x -> bin_start <= x < bin_end, eventlist.times)
+
+        if err_method == :poisson
```
matteobachetti, line 60: If the error is thrown before at line 51, this code should never execute

test/test_lightcurve.jl
```diff
@@ -0,0 +1,34 @@
...
+
+    @test isfile(sample_file)
+
+    # Test reading the sample file
+    eventlist = readevents(sample_file)
+    @test eventlist.filename == sample_file
+    @test length(eventlist.times) == length(times)
+    @test length(eventlist.energies) == length(energies)
+
+    # Test different bin sizes
```
matteobachetti, line 25: Since you are assigning exact event times, you should also be able to test that the counts in the light curve are exactly the ones you'd expect, rather than a generic >= 0

matteobachetti (changes requested):
Thanks @JawadAhmadCS! 
Some comments on the current implementation

src/lightcurve.jl
JawadAhmadCS: yes in we can use fit(Histogram, data, bins) from StatsBase.jl for efficient histogram binning. It is optimized and runs faster than manual looping.

JawadAhmadCS: I've implemented this. you can check it now.

test/test_lightcurve.jl
JawadAhmadCS: i have added please check now

src/lightcurve.jl
JawadAhmadCS: you are right but in my latest implementation i moved to using fit(Histogram, eventlist.times, bins), which simplifies the binning process and ensures correctness.

src/lightcurve.jl
```diff
@@ -0,0 +1,48 @@
...
+- The function bins event times into intervals of `bin_size`.
+- Errors are estimated using Poisson statistics.
+"""
+function create_lightcurve(eventlist::EventList{T}, bin_size::T; err_method::Symbol=:poisson) where T
+    min_time = minimum(eventlist.times)
+    max_time = maximum(eventlist.times)
+    bins = min_time:bin_size:max_time
+    hist = fit(Histogram, eventlist.times, bins)
+    counts = hist.weights
+    errors = sqrt.(counts)
```
fjebaker, line 46: You probably want to check that the `err_method` is valid (i.e. is `:poisson`), else the caller could pass anything and the code would not complain.

src/lightcurve.jl
```diff
@@ -0,0 +1,48 @@
...
+    return LightCurve{T}(bins[1:end-1], counts, errors, err_method)
```
fjebaker, line 47:
```suggestion
    return LightCurve{T}(bins, counts, errors, err_method)
```
Since the LightCurve will only be used for contiguous slices, we can use the standard that bins `i` and `i+1` mark the low and high time values.  It's important that we know what the width of the last time bin is.

test/test_lightcurve.jl
JawadAhmadCS: great catch i just updated the test to directly compare lightcurve.counts with expected values. appreciate the feedback

test/test_lightcurve.jl
```diff
@@ -0,0 +1,42 @@
...
+    
+        # Validate LightCurve properties
+        @test all(lightcurve.counts .>= 0)
+        @test lightcurve.err_method == :poisson
+        @test length(lightcurve.timebins) == length(lightcurve.counts)
+    
+        # Compute expected counts using Histogram directly
+        bins = minimum(eventlist.times):bin_size:maximum(eventlist.times)
+        expected_hist = fit(Histogram, eventlist.times, bins)
+        expected_counts = expected_hist.weights
```
fjebaker, line 37: So technically this is correct, but if `fit(Histogram, ...)` ever changed what it was doing, our results would be incorrect but the test would still pass. It would be preferred that you have a number of hard coded correct arrays that you comprae the counts to, e.g.

```julia
expected_counts = (
    [1, 2, 3, ...],
    [1, 2, 3, ...],
    # ...
)

bin_sizes = [1.0, 0.5, 2.0, 1.3]

for (bin_size, expected) in zip(bin_sizes, expected_counts)
    # ...
    @test lightcurve.counts == expected
end
```

fjebaker (changes requested):

test/test_lightcurve.jl
JawadAhmadCS: its done you can check now please review

src/lightcurve.jl
JawadAhmadCS: done.!

> Since the LightCurve will only be used for contiguous slices, we can use the standard that bins `i` and `i+1` mark the low and high time values. It's important that we know what the width of the last time bin is.

i have applied this change. now the bins include both start and end points, making it easier to determine the width of the last time bin. Thanks for the clarification!

JawadAhmadCS:
@matteobachetti @fjebaker please review i have resolve all..sorry for the delay in resolving this. I had a fever and some assignments to complete, so it took me a little longer.

JawadAhmadCS:
> Just a handful of little things, then this is ready to merge!

this is now a fully optimized solution.Please review @fjebaker @matteobachetti

kashish2210:
figuring out this pull request I have included the test cases and functions for example that can be implemented @JawadAhmadCS 
@matteobachetti and @fjebaker u guys can review this [notebook](https://github.com/kashish2210/notebooks/blob/main/events%20and%20create_lightcurve.ipynb)
further, some suggestions can be implemented :) and this can be used for advance implementation

src/lightcurve.jl
```diff
@@ -0,0 +1,51 @@
...
+
+- The function bins event times into intervals of `bin_size`.
+- Errors are estimated using Poisson statistics.
+"""
+function create_lightcurve(eventlist::EventList{T}, bin_size::T; err_method::Symbol=:poisson) where T
+    if err_method ∉ (:poisson,)
+        throw(ArgumentError("Invalid error method: $err_method. Supported methods: :poisson"))
+    end
+    min_time = minimum(eventlist.times)
+    max_time = maximum(eventlist.times)
```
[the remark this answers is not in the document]
JawadAhmadCS: its also done please review

fjebaker (approved):

src/lightcurve.jl
```diff
@@ -5,46 +5,70 @@ A structure representing a light curve, which is a time series of event counts.
...
+    
     hist = fit(Histogram, eventlist.times, bins)
-    counts = hist.weights
-    errors = sqrt.(counts)
+    counts = hist.weights 
+    
+    # to ensure counts vector has correct length
+    if length(counts) < length(bins) - 1
+        append!(counts, zeros(length(bins) - 1 - length(counts)))
+    end
```
fjebaker, line 69: I don't think this is a good idea. If counts didn't have the right length coming back from `fi(Histogram, ...)` that should be an error, but I don't think it's something we need to check. Does this actually ever trigger?

src/lightcurve.jl
```diff
@@ -5,46 +5,70 @@ A structure representing a light curve, which is a time series of event counts.
...
+- Ensures input validation for proper binning and error calculations.
+- Optimized to handle large event lists efficiently.
 """
 function create_lightcurve(eventlist::EventList{T}, bin_size::T; err_method::Symbol=:poisson) where T
     if err_method ∉ (:poisson,)
         throw(ArgumentError("Invalid error method: $err_method. Supported methods: :poisson"))
     end
-    min_time,max_time = extrema(eventlist.times)
+
+    isempty(eventlist.times) && throw(ArgumentError("Event list is empty. Cannot create a light curve."))
```
fjebaker, line 53: I don't think this is needed. If `times` is empty, I believe `fit(Histogram, ...)` just populates zeroes for all bins, which is perfectly fine.

src/lightcurve.jl
```diff
@@ -5,46 +5,70 @@ A structure representing a light curve, which is a time series of event counts.
...
+    bin_size <= 0 && throw(ArgumentError("Bin size must be positive."))
+    
+    min_time, max_time = extrema(eventlist.times)
     bins = min_time:bin_size:max_time
+    
+    if length(bins) < 2
+        push!(bins, max_time + bin_size)
+    end
```
fjebaker, line 61: This is a `MethodError`, you cannot push to a range. If there is only one bin, then `fit(Histogram, ...)` will return an empty array for the weights. I think that is fine as well, so this is unnecessary.

src/lightcurve.jl
```diff
@@ -5,46 +5,70 @@ A structure representing a light curve, which is a time series of event counts.
...
+    create_lightcurve(eventlist::EventList{T}, bin_size::T; err_method::Symbol=:poisson) where T
 
 Generate a light curve from an EventList by binning event times.
 
 ## Arguments
 
-- `eventlist::EventList{T}`: The input event list.
-- `bin_size::T`: The size of each time bin.
-- `err_method::Symbol`: The method for error estimation (`:poisson`).
+- eventlist::EventList{T}: The input event list.
```
fjebaker, line 32:
```suggestion
- `eventlist`: The input event list.
```
Docstrings should have things that refer to code in backticks, please add them back. You also don't need to annotate the type of parameters, since they are in the function signature.

Also, prefer using prose to describe the function and its arguments over lists. Have a look at some of the `Base` function documentation to get a feel for these things :+1:

src/lightcurve.jl
```diff
@@ -5,46 +5,70 @@ A structure representing a light curve, which is a time series of event counts.
...
+- Optimized to handle large event lists efficiently.
 """
 function create_lightcurve(eventlist::EventList{T}, bin_size::T; err_method::Symbol=:poisson) where T
     if err_method ∉ (:poisson,)
         throw(ArgumentError("Invalid error method: $err_method. Supported methods: :poisson"))
     end
-    min_time,max_time = extrema(eventlist.times)
+
+    isempty(eventlist.times) && throw(ArgumentError("Event list is empty. Cannot create a light curve."))
+    bin_size <= 0 && throw(ArgumentError("Bin size must be positive."))
```
fjebaker, line 54: Technically if you remove this then `bins` will just be length one (the first value), and the Histogram fit will be well defined and return an empty array. I think we can remove this too?

fjebaker (changes requested):
I @JawadAhmadCS  I can see that these changes are well intentioned, but they are in my opinion too defensive. Many of the checks you add are already handled by `fit(Histogram, ...)`, and allowing edge-cases like this can sometimes make someone's workflow easier. Since none of these return erroneous / wrong results, it's fine to just let them run through.

src/lightcurve.jl
JawadAhmadCS: You're right! fit(Histogram, ...) already handles this, so there's no need for the extra checks. I'll remove them

JawadAhmadCS: you are right! fit(Histogram, ...) already handles this, so there's no need. I'll remove it

src/lightcurve.jl
JawadAhmadCS: I've improved the docstring formatting.

src/lightcurve.jl
JawadAhmadCS: done

src/lightcurve.jl
JawadAhmadCS: got it! removing it since the histogram remains well-defined.

JawadAhmadCS:
please review @fjebaker 🤯

JawadAhmadCS:
Hi @fjebaker, I really appreciate your feedback! right now i am doing 3 internships along with university work, so it's a bit burden. but still trying to contribute whenever I can. by April 15, all my internships will be over, and I'll be totally free to contribute more actively.  i am sure you understand

fjebaker:
Hi @JawadAhmadCS no problem! I am swamped in my own work so have been using the odd bits of spare time to review your PRs. There's no urgency with this, so prioritize your other commitments and we can iterate slowly :)

JawadAhmadCS:
I think initial approved approach was correct, so I have reverted to that version. Please review and merge if everything looks good—it's fully okay.

JawadAhmadCS:
@fjebaker please review

fjebaker (approved):
Hi @JawadAhmadCS! Apologies it has taken a while to get round to finishing this review. This all looks good to me, I will ping @matteobachetti, as he needs to approve the changes before the merge conditions are met.

JawadAhmadCS:
@fjebaker  thanks a lot for the review and approval

JawadAhmadCS:
@matteobachetti whenever you get a chance, please take a look would love to get this merged

Issue #32, opened 2025-03-23
A GTI interface for the eventlists and lightcurves


fjebaker:
After #28, we need to start implementing some of the eventlist / lightcurve interfaces that stringray has. High up on that list is treating GTIs, and we want an equivalent of [`apply_gtis`](https://github.com/StingraySoftware/stingray/blob/a91c031b17d2e53f27594216f21f76b168fb316b/stingray/lightcurve.py#L1708). We can reuse the existing GTI Julia code that was implemented in the previous GSOC for much of this.

Unlike in the Python implementation, we will start by making this out-of-place only. That is, some functions with the following signatures:
```julia
apply_gtis(el::EventList, gtis) # should return an array of `EventList`, one for each GTI
apply_gtis(el::LightCurve, gtis) # should return an array of `LightCurve`, one for each GTI
```

Eventually, but not part of this issue, will be modifying things so we can use `@views` to filter the lightcurves, and also add the `inplace` variant, but that will require worrying about time bin continuity that is out of scope for this issue.

Then we also want something like [fill_bad_time_interval](https://github.com/StingraySoftware/stingray/blob/a91c031b17d2e53f27594216f21f76b168fb316b/stingray/base.py#L2204) for filling in gaps in the data.

```julia
fill_bad_time_intervals!(el::EventList, gtis; kwargs...) # should modify the given EventList inplace
```

Please describe how you are planning to implement these before you start 👍

JawadAhmadCS:
GTIs  are basically the time periods where our data is actually useful -no interruptions, no glitches, just clean, reliable observations. If we don’t handle these properly, we might end up using bad data, leading to wrong conclusions.

# How I Plan to Implement This:
`apply_gtis(el::EventList, gtis)` → Think of an event list like a long list of timestamps. We need to cut it into pieces, keeping only the parts that fall inside the GTIs. Each piece should still have its original metadata so we don’t lose important info.

`apply_gtis(el::LightCurve, gtis)` → A light curve is a bit trickier because it’s already split into bins. We have to carefully extract the valid sections while making sure the time bins stay continuous. Otherwise, we'd get weird gaps in the data.

`fill_bad_time_intervals!(el::EventList, gtis)` → Instead of just deleting the bad time intervals, sometimes we need to "fill in the blanks." This could mean interpolating missing data or adjusting timestamps intelligently.

fjebaker:
> Otherwise, we'd get weird gaps in the data.

[...]

PR #37, opened 2025-03-25
GTI Interface

JawadAhmadCS:
A GTI interface for the eventlists and lightcurves #32


JawadAhmadCS:
hi @fjebaker, I have implemented the gti interface for both eventlists and lightcurves. implementation includes two main functions:  `apply_gtis`, which filters data based on GTIs and returns separate objects while preserving metadata, and second `fill_bad_time_intervals`!, which modifies data in place to mask out periods outside gtis. it ensures proper input validation via check_gtis, correctly handles timing and energy data, and for lightcurves it zeroes out errors in bad intervals while keeping original values within gtis. please review

JawadAhmadCS:
hi @fjebaker I made some more changes. please review and let me know is i am on the right track. If yes, I’ll continue;If not, please suggest the correct approach

PR #53, opened 2025-07-02
adding interface method for gti

kashish2210:
### added some more fields in `fitsmetadata` like:
```julia
"Good Time Intervals (GTIs), if present in the file"
gti::Union{Nothing,Matrix{Float64}}
```
```julia
"GTI HDU name or index used"
gti_source::Union{Nothing,String,Int}
```
### and added read_gti_from_fits:
```julia
function read_gti_from_fits(fits_file::String; 
                           gti_hdu_candidates::Vector{String} = ["GTI", "STDGTI"],
                           gti_hdu_indices::Vector{Int} = Int[])::Tuple{Union{Nothing,Matrix{Float64}}, Union{Nothing,String}}
```
### added some interface using this and a utility function 
- majorly updated test cases for GTI

I have seprated this PR from #52 
since this will have updates in test and all so aboid much load 

tagging @matteobachetti  @stefanocovino @fjebaker for review :)


kashish2210:
#32

kashish2210:
taggin @fjebaker @stefanocovino  for review

src/utils.jl
```diff
@@ -27,3 +27,42 @@ function contiguous_regions(condition::AbstractVector{Bool})
     # Reshape the result into two columns
     return reshape(idx,2,length(idx) ÷ 2)'
 end
+#UTIL function used in recpies
+"""
+merge_overlapping_gtis(gtis::Matrix{Float64})::Matrix{Float64}
+Merge overlapping Good Time Intervals (GTIs).
+This function takes a matrix of GTIs, where each row represents a start and stop time,
```
fjebaker, line 34:
```suggestion
    merge_overlapping_gtis(gtis::Matrix{Float64})::Matrix{Float64}
   
Merge overlapping Good Time Intervals (GTIs).
This function takes a matrix of GTIs, where each row represents a start and stop time,
```

src/utils.jl
```diff
@@ -27,3 +27,42 @@ function contiguous_regions(condition::AbstractVector{Bool})
...
+If the input matrix has one or zero rows, it is returned unchanged.
+The function sorts the GTIs by start time, then iterates through them, merging any overlapping intervals.
+The output is a matrix with the same two-column format, containing the merged GTIs.
+"""
+function merge_overlapping_gtis(gtis::Matrix{Float64})::Matrix{Float64}
+    if size(gtis, 1) <= 1
+        return gtis
+    end
+    sort_indices = sortperm(gtis[:, 1])
+    sorted_gtis = gtis[sort_indices, :]
```
fjebaker, line 46: You can likely use `Base.permutecols!` or similar for this to avoid creating a copy of all of the GTIs.

src/utils.jl
```diff
@@ -27,3 +27,42 @@ function contiguous_regions(condition::AbstractVector{Bool})
...
+    current_stop = sorted_gtis[1, 2]
+    
+    for i in 2:size(sorted_gtis, 1)
+        start_time = sorted_gtis[i, 1]
+        stop_time = sorted_gtis[i, 2]
+        if start_time <= current_stop + 1e-6  # Small tolerance for floating point
+            # Merge intervals - extend the current stop time
+            current_stop = max(current_stop, stop_time)
+        else
+            merged = vcat(merged, [current_start current_stop])
```
fjebaker, line 59: Perhaps consider pre-allocating the `merged` array and assigning into it? This would avoid having to dynamically resize the array.

fjebaker (commented):
This is great! Just a handful of small comments.

src/utils.jl
kashish2210: done

kashish2210:
@matteobachetti I have added some missing functions from Stingray and their test cases: 

1. split_gtis_by_exposure ,
2. split_gtis_at_indices ,
3. merge_gtis,cross_gtis,
4. cross_two_gtis,
5. join_gtis,check_separate,
6. append_gtis ,
7. gti_border_bins ,
8. check_separate,
9. join_equal_gti_boundaries

your review required on this

kashish2210:
hey @matteobachetti 
so here I found out some problems while testing GTI with python versions i have tried to fix it 
### Problem 1
check_gtis incorrectly accepted empty GTI matrices like reshape(Float64[], 0, 2) due to all checks passing trivially:

- ndims(gti) == 2 && size(gti, 2) == 2 
- any(gti_end < gti_start) (false for empty arrays)
- any(@view(gti_start[begin+1:end]) < @view(gti_end[begin:end-1])) (false for empty views)

### Fixes I have made :

- Improved startbin validation: Changed from startbin < 1 to startbin < 0 to reflect 0-based indexing.
- Out-of-bounds skip: Added check to skip GTIs where startbin >= length(time), avoiding invalid access.
- Safe array access: Ensured startbin + 1 <= length(time) before reading time[startbin + 1].
- Consistent indexing: All bounds checks now consistently use 0-based logic for index calculations and 1-based access for arrays.

### Problem 2 
Problem in bin_intervals_from_gtis:

- When GTIs and time arrays don't overlap, the function raised a BoundsError instead of a more appropriate ArgumentError.

Cause I predicted:

- The function attempted to access indices before checking for overlaps.

Solution:

- Check for any overlap using extrema() early in the function.
- Raise a descriptive ArgumentError when no overlap exists.
- Added bounds checks to prevent out-of-range indexing.

### problem 3

- Problem in operations_on_gtis:
- Intersection of non-overlapping GTIs returned the original input GTI instead of an empty result.

Cause I predicted :

- The function directly returned the non-empty GTI without computing the actual operation when only one GTI was present.

Previous Approach was :

- Tried to patch with special-case handling (inefficient and non-general).

Solution I have implemented:

- Convert all GTIs (including empty ones) to IntervalSet objects uniformly.
- Apply the desired operation (intersect, union, etc.) across all sets.
- Let IntervalSets.jl handle the math robustly.
- Convert the result back to a matrix format.

I HAVE ADDED SOME EDGE TEST CASES TO :)
PLEASE NOTIFY ME THAT AM I ON RIGHTRACK OR NOT ?

matteobachetti (approved):

Issue #33, opened 2025-03-23
Fourier transforms and spectra


fjebaker:
Given an event list or light curve, we want to be able to compute the Fourier transform and obtain the spectra. Specifically, we want the [`AveragedPowerspectrum`](https://github.com/StingraySoftware/stingray/blob/a91c031b17d2e53f27594216f21f76b168fb316b/stingray/powerspectrum.py#L786).

Much of this code already exists in the Stringray.jl implementation, but need to be cleaned up / rewritten for the new primitives we are working with.

We minimally need something like the [`from_events`](https://github.com/StingraySoftware/stingray/blob/a91c031b17d2e53f27594216f21f76b168fb316b/stingray/powerspectrum.py#L466) constructor to construct an averaged power spectrum from the `EventList`. This constructor does not need all of the Python functionality, e.g. GTI treatment, as that will be handled separately for now.

We want something like:
```julia
struct AveragedPowerspectrum{T}
    freqs::Vector{T}
    power::Vector{T}
    power_errors::Vector{T}
end
```

I don't think any of the other attributes are relevant for now!

fjebaker:
Note: this is an involved task. It will involve implementing something like [this Python function](https://github.com/StingraySoftware/stingray/blob/a91c031b17d2e53f27594216f21f76b168fb316b/stingray/fourier.py#L1589), which has already been [implemented in Julia](https://github.com/StingraySoftware/Stingray.jl/blob/b29129ca322a2eca946327e7851c2784452d0d0d/src/fourier.jl#L227), but needs revisiting.

The temptation here is to write a lot of code to handle all of the different normalisations. I would urge you to avoid doing that, and instead provide the means by which we can add the other normalisations / features later, but provide for now just the default cases. This will keep the code changes short and manageable whilst we get the core of the functions right, and in doing so avoid having to rewrite a whole load of cases whilst we refine the solution.

JawadAhmadCS:
hi @fjebaker  It would be appreciated if you could provide a bit of the workflow to get started!. I’m thinking of working on this task.

kashish2210:
<img src="https://github.com/user-attachments/assets/32be4ce1-2e12-4612-9fde-4e37c8a37a6c" width="500" />


@fjebaker This going to my workflow suggest necessary things if @JawadAhmadCS is ohkay with this then assign this to me 

my [workflow](https://github.com/kashish2210/notebooks/blob/main/paridogarm%20(2).ipynb)

[...]

PR #58, opened 2025-07-28
Spectra 

kashish2210:
### so now we can do :
```julia
events = readevents("ni1200120104_0mpu7_cl.evt", load_gti=true, sort=true)
lc = create_lightcurve(events, 0.01)
# Calculate PDS from events with 100-second segments
segment_size = 100.0  # seconds
dt = 0.01  # time resolution
pds_result = avg_pds_from_eventlist(events, segment_size, dt; 
                                   norm="frac", 
                                   use_common_mean=true, 
                                   silent=false)
```
Output:
image

```julia
pds_result = avg_pds_from_lightcurve(lc, 100.0;  # 100-second segments
                                    norm="frac", 
                                    use_common_mean=true, 
                                    silent=false)
```
output
image

```julia
events2 = readevents("xmm_test.fits", load_gti=true, sort=true)
cs_result = avg_cs_from_eventlists(events, events2, segment_size, dt;
                                  norm="frac",
                                  use_common_mean=true,
                                  power_type="all")
lc2 = create_lightcurve(events2, 0.01)
cs_result = avg_cs_from_lightcurves(lc, lc2, 100.0;
                                   norm="frac",
                                   use_common_mean=true,
                                   power_type="all")
```
Output:
image

```julia
# some plotting examples, later I will put this in recipes
plot(pds_result.freq[subsample], pds_result.power[subsample],
      xscale=:log10, yscale=:log10,
      xlabel="Frequency (Hz)", ylabel="Power",
      title="Full PDS (Log-Log)", linewidth=1,
      label="PDS")
plot(histogram(log10.(pds_result.power), bins=100,
                      xlabel="log10(Power)", ylabel="Count",
                      title="Power Distribution", alpha=0.7,
                      label="Power Hist"))
```
Outputs:
image

image

### For a complete analysis, you may visit this [notebook](https://github.com/kashish2210/notebooks/blob/main/spectra%20(2).ipynb)

tagging @matteobachetti @fjebaker @stefanocovino for reviews


kashish2210:
#33

matteobachetti (changes requested):
At the very least, there should be a test for every new function.

I suggest a test that verifies that the new spectra function from event lists give the exact same result as the corresponding functions that get the times instead. 

In practice we should get something like
```spectrum_fun(events) .== spectrum_fun(events.time)```

File: src/fourier.jl [hunk 1]
```diff
@@ -674,17 +674,115 @@ function avg_cs_from_iterables(
...
+    flux_iterable2 = [counts2]
+    
+    return avg_cs_from_iterables(flux_iterable1, flux_iterable2, dt1; 
+                                norm=norm, use_common_mean=use_common_mean, 
+                                fullspec=fullspec, silent=silent, 
+                                power_type=power_type, return_auxil=return_auxil)
+end
+
+# Original functions (kept for backward compatibility)
+function avg_pds_from_events(times::AbstractVector{<:Real}, gti::AbstractMatrix{<:Real},
```
kashish2210, line 781: this functions

kashish2210:
I have added test cases @matteobachetti :)
I upgraded the format to make it simpler and compatible with Julia, as I have seen this type of format in @fjebaker repo
One more thing, I want to ask, I have kept [these previous functions](https://github.com/StingraySoftware/Stingray.jl/pull/58/files#r2257954547), shall I remove or just keep them? I thought it may help in vector time

PR #73, opened 2026-02-23
Replaced three @resumable functions with native Julia iterators:

kashish2210:
`generate_indices_of_segment_boundaries_unbinned` and `generate_indices_of_segment_boundaries_binned` in `gti.jl `now return zip and generator expressions instead of using @yield

`get_flux_iterable_from_segments` in `fourier.jl` now returns a generator expression instead of using @resumable with @yield


kashish2210:
@matteobachetti and @fjebaker, I have made some updates in gti.jl and fourier.jl, as we had expected this resumable function failure earlier; now it is compatible with Julia {newer version}{1.12}

kashish2210:
You can see the test failure in #74 for the @resumable function

matteobachetti:
@kashish2210 can you add a test case with Julia 1.12, so that you can test that indeed it resolves the issue in #74?

kashish2210:
> @kashish2210 can you add a test case with Julia 1.12, so that you can test that indeed it resolves the issue in #74?

yeah it resolves the issue

kashish2210:
@matteobachetti, what happened exactly `@resumable` macro broke in Julia 1.12 because Julia's internal compiler APIs changed. The package wasn't compatible with the newer Julia version, causing all tests to fail with MethodError.

PR #74, opened 2026-02-23
updating workflow julia version

kashish2210:
This is a small change to the Julia version from `1.11` to `1.12` in the workflow patch 
I have tested this locally. This workflow looks fine to me after merging #73. We can proceed with this


kashish2210:
@matteobachetti  and @fjebaker, may u can review this;)

PR #76, opened 2026-03-03
Add fftfreq import to runtests.jl

theralavineela:
The Fourier tests used `fftfreq` without explicitly importing the module
that defines it. With Julia ≥1.10 this results in:

    UndefVarError: fftfreq not defined

This PR:
- Imports `fftfreq` from AbstractFFTs in the test suite

This makes the dependency explicit and allows the test suite to run on
modern Julia versions without altering any functionality.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,5 +1,6 @@
 using Stingray
 using Test
+using AbstractFFTs: fftfreq 
 using FFTW, Distributions, Statistics, StatsBase, HDF5, FITSIO
 using Logging ,LinearAlgebra
 using CFITSIO


kashish2210:
Hey @theralavineela , thanks for your Pr You find this error in like test cases were failing ???
And in what exact version of Julia??

theralavineela:
Yes sir , I encountered the issue while running the test suite locally — some of the test cases were failing due to UndefVarError: fftfreq not defined. After investigating, I noticed that fftfreq was being used without explicitly importing it.

I reproduced this while running the tests on Julia v1.12.5. Adding the explicit import from AbstractFFTs resolved the issue locally and allowed the tests to run.

Please let me know if you would like me to test this on any other Julia versions as well.

kashish2210:
Oh ohaky may be we can consider this as refactor along with once ci in #74 will be updated we can test this also along with #73

theralavineela:
yeah sure sir ,thank you for the clarification :)

PR #79, opened 2026-05-15
feat: implement Power Colors module

Omiii-215:
## Implement Power Colors Module

This PR implements the Power Colors feature ported from the original Python Stingray package to `Stingray.jl`, following the methodology of Heil et al. (2015). 

The PR has been fully restructured into clean, granular commits to isolate the feature development, and heavily refactored.

### Summary of Changes

- **Core Computation (`power_colors.jl`)**: Added `power_color`, `hue_from_power_color`, and configuration defaults. Refactored to use idiomatic Julia (no explicit `Float64` casts, letting the compiler infer types from inputs, and avoiding unnecessary `local` declarations).
- **Spectral Integration (`fourier.jl`)**: Added `integrate_power_in_frequency_range` to support partial-bin integrations with Poisson noise subtraction. Optimized using `@views` for zero-allocation array slicing.
- **Dual Plotting Ecosystem**: 
  - Added publication-quality visualization functions (`plot_power_colors`, `plot_hues`) using `CairoMakie`.
  - Added lightweight `RecipesBase` wrapper types (`PowerColorPlot`, `HuePlot`) for exploratory data analysis across the broader `Plots.jl` ecosystem.
- **Strengthened Testing**: Achieved comprehensive test coverage (66 tests). Tests now rigorously verify the numerical correctness of computed scatter positions and angles, rather than just checking returned figure types.

*(Note: The previously included bug fixes for `fftfreq` and histograms have been reverted from this branch and will be submitted in a separate PR


src/Stingray.jl
```diff
@@ -6,6 +6,7 @@ using ProgressBars: tqdm as show_progress
 using DocStringExtensions
 using LinearAlgebra
 using Random
+using CairoMakie
```
kashish2210: I really like the idea of using cariomakie, but I will say you can use plot.JL it would be better for That would be better for exploratory data analysis, and you can even create your macro functions using recipe base and you can combine it with plot.JL

src/fourier.jl
```diff
@@ -759,3 +760,90 @@ function avg_cs_from_events(times1:: AbstractVector{<:Real}, times2:: AbstractVe
...
+    if isnothing(df)
+        df_arr = fill(median(diff(freq)), length(freq))
+    elseif df isa Real
+        df_arr = fill(Float64(df), length(freq))
+    else
+        df_arr = Float64.(df)
+    end
+
+    # Build frequency mask: bins that overlap with the requested range
+    frequency_mask = ((freq .+ df_arr ./ 2) .> frequency_range[1]) .&
```
kashish2210: This function is very crucial, and you have to take care of BIN edges here this can be written like this
`frequency_mask = (freq .+ dfs./2 .> frequency_range[1]) .& (freq .- dfs./2 .< frequency_range[2])`

As of ur current stimulation uses center only we will miss the edges

src/fourier.jl
```diff
@@ -759,3 +760,90 @@ function avg_cs_from_events(times1:: AbstractVector{<:Real}, times2:: AbstractVe
...
+    correction_ratios[1] = (freqs_to_integrate[1] + 0.5 * dfs_to_integrate[1] -
+                            frequency_range[1]) / dfs_to_integrate[1]
+    correction_ratios[end] = (frequency_range[2] - freqs_to_integrate[end] +
+                              0.5 * dfs_to_integrate[end]) / dfs_to_integrate[end]
+    dfs_to_integrate = dfs_to_integrate .* correction_ratios
+
+    powers_to_integrate = pow[frequency_mask]
+
+    if isnothing(power_err)
+        power_err_to_integrate = real.(powers_to_integrate) ./ sqrt(m)
```
kashish2210: I really appreciate using real here good!

src/fourier.jl
```diff
@@ -759,3 +760,90 @@ function avg_cs_from_events(times1:: AbstractVector{<:Real}, times2:: AbstractVe
...
+    freqs_to_integrate = freq[frequency_mask]
+    pois_to_integrate = pois[frequency_mask]
+    dfs_to_integrate = df_arr[frequency_mask]
+
+    # Correction ratios for partial bins at the edges
+    correction_ratios = ones(length(freqs_to_integrate))
+    correction_ratios[1] = (freqs_to_integrate[1] + 0.5 * dfs_to_integrate[1] -
+                            frequency_range[1]) / dfs_to_integrate[1]
+    correction_ratios[end] = (frequency_range[2] - freqs_to_integrate[end] +
+                              0.5 * dfs_to_integrate[end]) / dfs_to_integrate[end]
```
kashish2210: I am not sure about sign in here : 

The first bin's correction ratio should be the fraction of the bin that lies within the range, i.e. (upper_edge_of_bin - range_start) / bin_width. The minus sign on 0.5 * df makes it compute the lower edge instead, which gives a ratio > 1 or negative in many cases — both physically wrong.

U may look at this I made some correctness 
`correction_ratios[1] = (freqs_to_integrate[1] + 0.5 * dfs_to_integrate[1] - frequency_range[1]) / dfs_to_integrate[1] `

kashish2210 (commented):
Hey @Omiii-215 thats a good pr I have given some of the feedback's u may review!:)

src/power_colors.jl
```diff
@@ -0,0 +1,566 @@
...
+# Returns
+- `hue`: Angle(s) in radians.
+
+# References
+Heil et al. 2015, MNRAS, 448, 3348.
+"""
+function hue_from_power_color(pc0, pc1; center = (4.51920, 0.453724))
+    log_pc0 = log10.(pc0)
+    log_pc1 = log10.(pc1)
+    log_center = log10.([center[1], center[2]])
```
fjebaker:
```suggestion
    log_center = log10.(center)
```

src/fourier.jl
```diff
@@ -759,3 +760,90 @@ function avg_cs_from_events(times1:: AbstractVector{<:Real}, times2:: AbstractVe
...
+function integrate_power_in_frequency_range(
+    frequency::AbstractVector{<:Real},
+    power::AbstractVector,
+    frequency_range;
+    power_err::Union{Nothing, AbstractVector} = nothing,
+    df::Union{Nothing, Real, AbstractVector} = nothing,
+    m::Int = 1,
+    poisson_power::Union{Real, AbstractVector} = 0,
+)
+    freq = Float64.(frequency)
```
fjebaker: I don't know if I understand what these `Float64` casts are doing? In general with Julia, it's best to not refer to concrete primatives, but to let the compiler resolve the types.

src/fourier.jl
```diff
@@ -759,3 +760,90 @@ function avg_cs_from_events(times1:: AbstractVector{<:Real}, times2:: AbstractVe
...
+        df_arr = Float64.(df)
+    end
+
+    # Build frequency mask: bins that overlap with the requested range
+    frequency_mask = ((freq .+ df_arr ./ 2) .> frequency_range[1]) .&
+                     ((freq .- df_arr ./ 2) .< frequency_range[2])
+
+    freqs_to_integrate = freq[frequency_mask]
+    pois_to_integrate = pois[frequency_mask]
+    dfs_to_integrate = df_arr[frequency_mask]
```
fjebaker: To avoid allocating, you could also use `@views` here. Since the mask will be a contiguous slice, it should be cache friendly also.

src/fourier.jl
fjebaker: As Om has done it here, it is consistent with the Python package.

src/power_colors.jl
```diff
@@ -0,0 +1,566 @@
...
+    # Validate freq_edges
+    if length(freq_edges) != 5
+        throw(ArgumentError("freq_edges must have 5 elements"))
+    end
+
+    freq = Float64.(frequency)
+    pow = Float64.(power)
+
+    # Compute df if not provided
+    local df_val::Float64
```
fjebaker: `local` keyword not needed. Then, as mentioned in a previous comment, derive the types from the input arguments:
```suggestion
    df_val = zero(eltype(freq))
```

src/power_colors.jl
```diff
@@ -0,0 +1,566 @@
...
+
+    if minimum(freq_edges) < input_frequency_low_edges[1]
+        throw(ArgumentError("The minimum frequency is larger than the first frequency edge"))
+    end
+    if maximum(freq_edges) > input_frequency_high_edges[end]
+        throw(ArgumentError("The maximum frequency is lower than the last frequency edge"))
+    end
+
+    # Default error
+    local pow_err::Vector{Float64}
```
fjebaker: Same as above.

test/test_power_colors.jl
```diff
@@ -0,0 +1,146 @@
...
+    end
+
+
+    @testset "plot_hues: runs without error" begin
+        for plot_spans in [true, false]
+            for polar in [true, false]
+                fig, ax = plot_hues(rms_val, rmse_val, pc0, pc1;
+                    polar = polar, plot_spans = plot_spans,
+                    configuration = configuration)
+                @test fig isa CairoMakie.Figure
```
fjebaker: Okay but this test doesn't really help much. It would be better to check that the values that it's going to plot are correct. We could delete the whole implementation of `plot_hues` and it would still pass this test provided it returns an empty Figure.

fjebaker (changes requested):
Hello @Omiii-215 This is cool! I've got a few comments are some requested changes. Then, as we discussed during the call earlier, see if you can break this commit and PR up into a fix plus a feature commit and PR.

If you need any help, post in the Slack and Matteo or I will try to get back to you as soon as possible :)

src/fourier.jl
kashish2210: As fergus said Considering it's is an consistent with python package then we might ignore this comment

Omiii-215:
**Power Colors with Spans**

Screenshot 2026-05-26 at 1 04 37 PM

**Hue vs RMS (Polar)**
Screenshot 2026-05-26 at 1 06 03 PM

Omiii-215:
@fjebaker @matteobachetti  **Hue vs RMS (Cartesian)**
Screenshot 2026-05-27 at 1 44 23 AM

I do think the Plots came out pretty well :)

Omiii-215:
@kashish2210 Also I did try using recipe base and kinda refactored some things

kashish2210:
yup that's looks nice @Omiii-215 good work

src/power_colors.jl
```diff
@@ -0,0 +1,551 @@
...
+Wrapper type for plotting hue vs rms via `Plots.jl`.
+
+# Usage
+```julia
+using Plots
+plot(HuePlot(rms, rms_err, pc0, pc1))
+plot(HuePlot(rms, rms_err, pc0, pc1; plot_spans=true))
+```
+"""
+struct HuePlot
```
kashish2210, line 506: Maybe be for plotting u can have a Plot folder inside it u can keep all the recipes like recipe_power_colors.jl
it was a small suggestion by @matteobachetti  earlier when I was doing

kashish2210 (commented):
you have implemented it very well!

Project.toml
```diff
@@ -5,38 +5,44 @@ version = "0.1.0"
 
 [deps]
 CFITSIO = "3b1b4be9-1499-4b22-8d78-7db3344d1961"
+CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
```
fjebaker: Avoid depending on a particular backend of Makie.jl. Let a user decide which backend they want to use, and instead depending on the interface library Makie.jl itself. For some examples, see https://docs.makie.org/stable/explanations/recipes.html

By contrast, it's fine for the test suite to depend on CairoMakie.jl, as a user of the library does not need to install the test dependencies.

fjebaker (commented):
Really nice work, and thanks for splitting the commits up.

My only remaining comment is regarding the CairoMakie.jl dependency. See below:

Project.toml
Omiii-215: @fjebaker I'll updated Project.toml to:

- Replace CairoMakie with Makie in [deps] and [compat].
- Move CairoMakie into the [extras] and [targets] sections so that it's correctly used as a test-only dependency.

src/power_colors.jl
fjebaker: That's a good point! We'll not block this PR any further though and refactor where the code lives in a subsequent PR.

fjebaker (approved):

PR #80, opened 2026-05-28
fix: correct fftfreq sampling rate and histogram bin edges

Omiii-215:
Fix two bugs in the fourier module:

1. fftfreq(n_bin, dt) → fftfreq(n_bin, 1/dt): Julia's FFTW.fftfreq expects the sampling rate (1/dt), not the sample spacing (dt). The Python convention (numpy.fft.fftfreq) is the opposite, which caused incorrect frequency arrays when porting.

2. fit(Histogram, ...; nbins=n_bin) → fit(Histogram, ..., edges): Using nbins does not guarantee bin edges align with segment boundaries. Explicit edges via range(0, stop=e-s, length=n_bin+1) ensure deterministic, gap-free binning of unbinned event data.

### DIFF
--- a/test/test_fourier.jl
+++ b/test/test_fourier.jl
@@ -52,6 +52,40 @@ end
     @test filter(x -> x >0, freq) == freq[goodbins]
 end
 
+@testset "test_fftfreq_uses_sampling_rate" begin
+    dt = 0.5
+    segment_size = 50.0
+    n_bin = round(Int, segment_size / dt)
+    times = sort(rand(Uniform(0, 100), 10000))
+    gti = [[0 100];;]
+    result = avg_pds_from_events(times, gti, segment_size, dt, silent=true)
+    freqs = result[!, "freq"]
+    df = 1 / segment_size
+    @test freqs[1] ≈ df
+    @test (freqs[2] - freqs[1]) ≈ df
+    @test maximum(freqs) ≈ (n_bin ÷ 2 - 1) * df
+end
+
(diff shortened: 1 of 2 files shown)


fjebaker (approved):

fjebaker:
This looks great! Could you add a test that checks the output of these two things, so that if they were ever regressed the tests fail.

Omiii-215:
Committed the tests @fjebaker 👍🏻

test/test_fourier.jl
```diff
@@ -52,6 +52,38 @@ end
     @test filter(x -> x >0, freq) == freq[goodbins]
 end
 
+@testset "test_fftfreq_uses_sampling_rate" begin
+    n = 100
+    dt = 0.5
+    freq = fftfreq(n, 1/dt)
```
fjebaker: Okay, but this test is not testing the Stingray.jl code, only the FFTW library. We could undo your fix commit and this test would still pass. Make sure you're testing the Stringray.jl library and not one of the dependencies (~;

Omiii-215: Thank you! Will be Doing that !

Omiii-215:
**Now both tests properly exercise Stingray.jl code**

- fftfreq test → calls avg_pds_from_events, checks returned frequencies
- histogram test → calls get_flux_iterable_from_segments, checks no events are dropped (sum(flux) == expected_count)

fjebaker (approved):

PR #81, opened 2026-06-14
Core spectral types

Omiii-215:
### Key Changes
- **Core Structs**: Added `AbstractCrossspectrum`, `Crossspectrum`, `AbstractPowerspectrum`, and `Powerspectrum`.
- **Constructors**: Implemented high-level constructors allowing these structs to be initialized directly from `EventList` and `LightCurve` objects.
- **Independent Metadata Handling**: Temporarily bypasses the `fourier.jl` DataFrame output for metadata calculations. Instead, it computes statistically accurate metadata (`m` segments, `df`, `nphots`, Poisson errors) directly inside the constructors to prevent known DataFrame edge-case bugs. 
- **Testing**: Added comprehensive test suites (`test_crossspectrum.jl` and `test_powerspectrum.jl`) verifying struct fields and normalizations (including Leahy Parseval verification and Fractional RMS limits). All tests pass.

### DIFF
--- a/src/crossspectrum.jl
+++ b/src/crossspectrum.jl
@@ -101,6 +101,77 @@ function _count_valid_segments(gti::Matrix, segment_size::Real)
 end
 
 """
+    _compute_spectral_params(segment_size, dt, gti) -> NamedTuple
+
+Compute shared spectral metadata (n_bin, adjusted dt, df, m) from segment
+parameters and GTIs. Used internally by `Crossspectrum` and `Powerspectrum`
+constructors.
+"""
+function _compute_spectral_params(segment_size::Real, dt::Real, gti::Matrix)
+    n_bin = round(Int, segment_size / dt)
+    dt_adj = segment_size / n_bin
+    df = 1.0 / segment_size
+    m = _count_valid_segments(gti, segment_size)
+    return (; n_bin, dt=dt_adj, df, m)
+end
+
+"""
+    _extract_spectral_result(result; complex_power=true) -> NamedTuple
+
+Extract freq, power, unnorm_power, and auxiliary PDS columns from a DataFrame
+returned by `avg_cs_from_events` or `avg_pds_from_events`. Throws if `result`
+is `nothing`.  Set `complex_power=false` for real-valued power (Powerspectrum).
+"""
+function _extract_spectral_result(result; complex_power::Bool=true)
+    if isnothing(result)
+        throw(ArgumentError("No valid segments found. Check GTIs and segment_size."))
+    end
+
+    freq = Float64.(result[!, "freq"])
+    if complex_power
+        power = Complex{Float64}.(result[!, "power"])
+        unnorm_power = Complex{Float64}.(result[!, "unnorm_power"])
+    else
+        power = Float64.(real.(result[!, "power"]))
+        unnorm_power = Float64.(real.(result[!, "unnorm_power"]))
+    end
+
+    pds1 = "pds1" in names(result) ? Float64.(real.(result[!, "pds1"])) : nothing
+    pds2 = "pds2" in names(result) ? Float64.(real.(result[!, "pds2"])) : nothing
+
+    return (; freq, power, unnorm_power, pds1, pds2)
+end
+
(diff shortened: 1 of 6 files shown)


Omiii-215:
Hi @fjebaker @matteobachetti 

I have implemented the core spectral data structures (`AbstractCrossspectrum`, `Crossspectrum`, `AbstractPowerspectrum`, `Powerspectrum`) matching Python Stingray, along with high-level constructors for `EventList` and `LightCurve` objects. I've also added comprehensive test suites verifying their accuracy (including Leahy Poisson noise verification and Fractional RMS/Parseval's theorem checks).

 issue: The constructors currently wrap the existing `avg_pds_from_events` and `avg_cs_from_events` functions in `fourier.jl`. However, relying solely on the generic `DataFrame` output from these functions is unreliable because it occasionally omits or miscalculates critical statistical metadata like `m`, `df`, and `nphots` (which led to some edge cases around segment extraction).

To keep this PR tightly scoped and avoid breaking internal code, my current approach was to use `fourier.jl` purely as a math engine for the FFTs. Inside the new constructors, I wrote robust logic to independently calculate the proper metadata directly from the input data to ensure the structs are scientifically accurate now.

Since stripping out the DataFrame outputs and fixing the internal segment logic in `fourier.jl` will be a significant refactor, does it make sense to leave my current approach in this PR and address the `fourier.jl` refactor in a separate PR?

Omiii-215:
Also next will be Building upon the core spectral types we established in PR 1 (`Crossspectrum` and `Powerspectrum`), Will focus entirely on **frequency rebinning and RMS calculations**. :

- **Core Rebinning **: Implement `rebin_data` (linear) and `rebin_data_log` (logarithmic).

- **Linear Rebinning (`rebin`)**: Add methods to our new spectral structs that let users combine frequency bins to improve signal-to-noise ratios, while automatically updating statistical metadata like `m`.

- **Logarithmic Rebinning (`rebin_log`)**: Add methods to safely group high-frequency bins geometrically, using a tracking vector (`k`) to log exactly how many original bins were merged.

- **Fractional RMS (`compute_rms`)**: Implement a function for `Powerspectrum` that integrates power over a specific frequency range, subtracts Poisson noise, and returns the exact fractional RMS amplitude.

- **Parity Testing**: Port over the full suite of Python tests to mathematically prove our rebinned spectra and RMS calculations behave identically to the original package.

src/crossspectrum.jl
```diff
@@ -0,0 +1,276 @@
...
+    unnorm_power = Complex{Float64}.(result[!, "unnorm_power"])
+
+    # Compute metadata from input parameters
+    n_bin = round(Int, segment_size / dt)
+    dt_adj = segment_size / n_bin
+    df_val = 1.0 / segment_size
+    m_val = _count_valid_segments(common_gti, segment_size)
+
+    power_err = power ./ sqrt(m_val)
+    unnorm_power_err = unnorm_power ./ sqrt(m_val)
```
fjebaker: This code is common between both `Crossspectrum` constructors, and largely the same in the `Powerspectrum` constructor. It would be cleaner to have a single version of all this metadata and normalisation code that gets called by each of the constructors.

src/crossspectrum.jl
```diff
@@ -0,0 +1,276 @@
...
+    # Extract auxiliary PDS if computed
+    _pds1 = "pds1" in names(result) ? Float64.(real.(result[!, "pds1"])) : nothing
+    _pds2 = "pds2" in names(result) ? Float64.(real.(result[!, "pds2"])) : nothing
+
+    # Compute photon statistics
+    nphots1_val = Float64(length(times(ev1))) / m_val
+    nphots2_val = Float64(length(times(ev2))) / m_val
+    nphots_val = sqrt(nphots1_val * nphots2_val)
+    countrate1_val = nphots1_val / segment_size
+    countrate2_val = nphots2_val / segment_size
```
fjebaker: This stuff too!

fjebaker (approved):
Great stuff @Omiii-215 !

